PR #55, opened 2021-01-31
Add optional linting rules for govcloud IAM policies

circa10a:
## Adding 3 new iam policies linting rules

These rules primarily apply to AWS GovCloud consumers. I used the word `friendly` as is does not mandate that terraform code only work for GovCloud, but to **forcibly enable support for it.**  which is explained with code in **why** section. Having this rules merged and implemented into this plugin would be a huge lift for my team when doing code reviews. 

1. `aws_iam_policy_document_gov_friendly`- [`aws_iam_policy_document` data source](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/iam_policy_document)
2. `aws_iam_policy_gov_friendly` - [`aws_iam_policy` resource](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/iam_policy)
3. `aws_iam_role_policy_gov_friendly` - [`aws_iam_role_policy` resource](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/iam_role_policy)

## Why

1. Some firms have strict requirements for what AWS resources have access to government accounts. Usually only resources within the `arn:aws-us-gov:` scope are allowed.
2. When developing reusable terraform modules for many AWS accounts, arn separators are usually converted into variables when creating resources in gov and non-gov accounts like so :

```hcl
locals {
  arn_sep = var.is_govcloud ? "aws-us-gov" : "aws"
}

resource "aws_iam_policy" "policy" {
  name        = "test_policy"
  path        = "/"
  description = "My test policy"
  policy      = <<EOF
  {
    "Version": "2012-10-17",
    "Statement": [
	  {
	    "Action": [
		  "ec2:Describe*"
	    ],
	    "Effect": "Allow",
	    "Resource": "arn:${local.arn_sep}:s3:::my_bucket/*""
	  }
    ]
  }
EOF
}
```


circa10a:
@wata727 Could you please review? 🙏

rules/aws_iam_policy_document_gov_friendly_arns.go
```diff
@@ -0,0 +1,75 @@
...
+// Link returns the rule reference link
+func (r *AwsIAMPolicyDocumentGovFriendlyArnsRule) Link() string {
+	return ""
+}
+
+// Check checks the pattern is valid
+func (r *AwsIAMPolicyDocumentGovFriendlyArnsRule) Check(runner tflint.Runner) error {
+	log.Printf("[TRACE] Check `%s` rule", r.Name())
+	return runner.WalkResourceBlocks(r.resourceType, statementBlockName, func(statementBlock *hcl.Block) error {
+		attributes, diags := statementBlock.Body.JustAttributes()
```
wata727, line 56: `JustAttributes` will fail when the `statement` block has a nested block. In this case, this is the `principals` block.
https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/iam_policy_document#principals

It is recommended to extract only the required attributes as shown below:

```suggestion
		content, _, diags := statementBlock.Body.PartialContent(&hcl.BodySchema{
			Attributes: []hcl.AttributeSchema{
				{Name: r.attributeName},
			},
		})
```

docs/rules/aws_iam_policy_document_gov_friendly.md
```diff
@@ -0,0 +1,79 @@
...
+```hcl
+rule "aws_iam_policy_document_gov_friendly" {
+  enabled = true
+}
+```
+
+## Examples
+
+```hcl
+resource "aws_iam_policy_document" "example" {
```
wata727, line 16: Probably `data` is correct :)

rules/aws_iam_policy_document_gov_friendly_arns.go
```diff
@@ -0,0 +1,75 @@
...
+}
+
+// Severity returns the rule severity
+func (r *AwsIAMPolicyDocumentGovFriendlyArnsRule) Severity() string {
+	return tflint.WARNING
+}
+
+// Link returns the rule reference link
+func (r *AwsIAMPolicyDocumentGovFriendlyArnsRule) Link() string {
+	return ""
```
wata727, line 49: For rules with documentation, it is recommended `Link` returns the URL so that the user can access the detailed explanation.

```suggestion
	return project.ReferenceLink(r.Name())
```

wata727:
Interesting. If you have a few rules about GovCloud, it also is interesting to cut them out into a dedicated ruleset (like tflint-ruleset-aws-govcloud) in the future.

rules/aws_iam_policy_document_gov_friendly_arns.go
circa10a: Great catch! Also added to tests.

docs/rules/aws_iam_policy_document_gov_friendly.md
circa10a: 🙌

circa10a:
> Interesting. If you have a few rules about GovCloud, it also is interesting to cut them out into a dedicated ruleset (like tflint-ruleset-aws-govcloud) in the future.

I can't see there be many just yet, but in the event that there are, probably a good idea :)

circa10a:
@wata727 any other concerns with this PR?

wata727 (approved):
Thanks 👏

circa10a:
@wata727 When did you plan on releasing the next version of this ruleset to be consumed by `tflint` ?

wata727:
Currently, we are planning to release it once a month except for bug fixes.

PatMyron:
In addition to AWS GovCloud, [also applies to AWS China](https://docs.aws.amazon.com/general/latest/gr/aws-arns-and-namespaces.html#arns-syntax) (and the [isolated partitions](https://docs.aws.amazon.com/sdk-for-go/api/aws/endpoints/#pkg-constants))
Could also avoid the [input variables](https://www.terraform.io/language/values/variables) by using the [`aws_partition` data source](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/partition)

PatMyron:
I'd also love for this to be a general rule, regardless of resource type. Best to avoid hardcoding `arn:aws:` everywhere, not just in these 3 resource types: https://github.com/aws-cloudformation/cfn-lint/pull/1805

Issue #60, opened 2021-02-07
Lint valid engine for RDS


chaspy:
https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/db_instance

I'd like to lint if `engine` is valid.
Can I try it?

circa10a:
Fixed in #61

PR #61, opened 2021-02-07
Add lint for db_instance's engine attribute

chaspy:
ref: https://github.com/terraform-linters/tflint-ruleset-aws/issues/60

### DIFF
--- a/docs/rules/README.md
+++ b/docs/rules/README.md
@@ -11,6 +11,7 @@ These rules warn of possible errors that can occur at `terraform apply`. Rules m
 |aws_alb_invalid_security_group|✔|
 |aws_alb_invalid_subnet|✔|
 |aws_db_instance_invalid_db_subnet_group|✔|
+|aws_db_instance_invalid_engine||
 |aws_db_instance_invalid_option_group|✔|
 |aws_db_instance_invalid_parameter_group|✔|
 |aws_db_instance_invalid_type||
@@ -52,6 +53,7 @@ These rules warn of possible errors that can occur at `terraform apply`. Rules m
 These rules suggest to better ways.
 
 - [aws_instance_previous_type](aws_instance_previous_type.md)
+- [aws_db_instance_invalid_engine](aws_db_instance_invalid_engine.md)
 - [aws_db_instance_previous_type](aws_db_instance_previous_type.md)
 - [aws_db_instance_default_parameter_group](aws_db_instance_default_parameter_group.md)
 - [aws_elasticache_cluster_previous_type](aws_elasticache_cluster_previous_type.md)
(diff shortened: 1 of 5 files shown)


chaspy:
@wata727 
Could you review this PR?
I'm afraid I don't still understand how the ruleset is used though, I wrote this with reference to `aws_db_instance_invalid_type.go`

rules/aws_db_instance_invalid_engine.go
```diff
@@ -0,0 +1,80 @@
...
+}
+
+// Severity returns the rule severity
+func (r *AwsDBInstanceInvalidEngineRule) Severity() string {
+	return tflint.ERROR
+}
+
+// Link returns the rule reference link
+func (r *AwsDBInstanceInvalidEngineRule) Link() string {
+	return project.ReferenceLink(r.Name())
```
wata727, line 60: This rule doesn't have documentation, so it's better to return an empty string instead.

By the way, the `aws_db_instance_invalid_type` rule also returns a link, but since there is no documentation, it seems that this should also return an empty string... 😓

wata727 (commented):
I left a suggestion but looks pretty good overall 👍

rules/aws_db_instance_invalid_engine.go
chaspy: Understand, cool 🆒 
https://github.com/terraform-linters/tflint-ruleset-aws/blob/master/project/main.go#L10
https://github.com/terraform-linters/tflint-ruleset-aws/tree/v0.2.1/docs/rules

I will add a document for the rule 👍

chaspy: Addressed 👍  https://github.com/terraform-linters/tflint-ruleset-aws/pull/61/commits/6a9575c5f3ed4e2b92fb230a5220433e11ece464 
How about this?

docs/rules/README.md
```diff
@@ -52,6 +53,7 @@ These rules warn of possible errors that can occur at `terraform apply`. Rules m
 These rules suggest to better ways.
 
 - [aws_instance_previous_type](aws_instance_previous_type.md)
+- [aws_db_instance_invalid_engine](aws_db_instance_invalid_engine.md)
```
wata727, line 56: It's listed above as possible errors, so we don't need to list it here.

chaspy: Addressed. https://github.com/terraform-linters/tflint-ruleset-aws/pull/61/commits/1a8a65629aaec98305a8f14438051660ff0d1f55

chaspy:
Ah, maybe need to add here? https://github.com/terraform-linters/tflint-ruleset-aws/blob/6a9575c5f3ed4e2b92fb230a5220433e11ece464/rules/provider.go#L9

wata727 (approved):
👍

PatMyron:
Wish these could [also be pulled](https://github.com/terraform-linters/tflint/pull/274) from the [SDK](https://github.com/aws/aws-sdk-go/blob/main/models/apis/rds/2014-10-31/api-2.json) :/

PR #88, opened 2021-03-26
Update link to full list of rules

william-richard:
The old link 404'd - looks like the repo was reorganized at some point and this link needs to be updated to match.


wata727:
Thanks!

PR #92, opened 2021-03-29
Update aws_mq_broker_invalid_engine_type.go

richt3211:
Adding support for RabbitMQ Engine type in aws mq. 

RabbitMQ was recently added as a supported engine type in AWS. Tflint throws an error when the engine type is set as "RabbitMQ", which is now a valid engine type in the aws terraform provider. 

This change allows for the use of the "RabbitMQ" engine type.


wata727 (approved):
👍

PR #99, opened 2021-04-10
add TFLINT_PLUGIN_DIR option to README

lachrimae:
Hey, I was working with this tool and was confused by the documentation. Upon reading this README, I thought that the two options described here were the only two available. I happened to be working in an environment where the `TFLINT_PLUGIN_DIR` option was the only one that would work. 

I eventually found out about the option from [these docs](https://github.com/terraform-linters/tflint/blob/master/docs/user-guide/config.md). I think listing all three plugin directory methods would be clearer and might save someone time down the road.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ This ruleset focus on possible errors and best practices about AWS resources. Ma
 
 ## Installation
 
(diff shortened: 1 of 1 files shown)


wata727:
Thanks!

Issue #124, opened 2021-06-08
M1 (arm64) release?


jayzes:
Hey there,

Trying to set up `tflint` with this ruleset on an M1 Mac and running into this error:

```
Installing `aws` plugin...
Failed to install a plugin. An error occurred:

Error: Failed to download tflint-ruleset-aws_darwin_arm64.zip: file not found in the GitHub release. Does the release contain the file with the correct name ?
```

I'm guessing this is because `goreleaser` / the latest release isn't set up to build arm64 by default, though looks like it is supported upstream now: https://github.com/goreleaser/goreleaser/issues/1952. Would you be open to a PR changing that config in the project?

wata727:
Sure!

wata727:
v0.4.3 has been released
https://github.com/terraform-linters/tflint-ruleset-aws/releases/tag/v0.4.3

PR #125, opened 2021-06-09
Update GoReleaser config to build arm64 for M1

jayzes:
This should solve https://github.com/terraform-linters/tflint-ruleset-aws/issues/124

While I'm in here, I also updated the actions config to use [the published GoReleaser action](https://github.com/goreleaser/goreleaser-action) instead of a custom script.


wata727:
Thanks!

Issue #140, opened 2021-07-16
aws_elasticache_replication_group same basic rules


FSJRay:
It looks like aws_elasticache_replication_group could use some of the rules as aws_elasticache_cluster such as node type, default parameter group, etc. Is this of interest to the community?

wata727:
Looks good 👍

PR #143, opened 2021-07-19
aws_elasticache_replication_group

Rihoj:
fixes terraform-linters/tflint-ruleset-aws#140

Adds rules for aws_elasticcache_replication_group resource

### DIFF
--- a/rules/aws_elasticache_replication_group_default_parameter_group.go
+++ b/rules/aws_elasticache_replication_group_default_parameter_group.go
@@ -43,7 +43,7 @@ func (r *AwsElastiCacheReplicationGroupDefaultParameterGroupRule) Link() string 
 	return project.ReferenceLink(r.Name())
 }
 
-var defaultElastiCacheParameterGroupRegexp = regexp.MustCompile("^default")
+var defaultElastiCacheReplicationParameterGroupRegexp = regexp.MustCompile("^default")
 
 // Check checks the parameter group name starts with `default`
 func (r *AwsElastiCacheReplicationGroupDefaultParameterGroupRule) Check(runner tflint.Runner) error {
--- a/rules/aws_elasticache_replication_group_default_parameter_group.go
+++ b/rules/aws_elasticache_replication_group_default_parameter_group.go
@@ -1,0 +1,65 @@
+package rules
+
+import (
+	"fmt"
+	"regexp"
+
+	hcl "github.com/hashicorp/hcl/v2"
+	"github.com/terraform-linters/tflint-plugin-sdk/tflint"
+	"github.com/terraform-linters/tflint-ruleset-aws/project"
+)
+
+// AwsElastiCacheReplicationGroupDefaultParameterGroupRule checks whether the cluster use default parameter group
+type AwsElastiCacheReplicationGroupDefaultParameterGroupRule struct {
+	resourceType  string
+	attributeName string
+}
+
+// NewAwsElastiCacheReplicationGroupDefaultParameterGroupRule returns new rule with default attributes
+func NewAwsElastiCacheReplicationGroupDefaultParameterGroupRule() *AwsElastiCacheReplicationGroupDefaultParameterGroupRule {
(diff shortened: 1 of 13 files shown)


wata727 (commented):
Good job!

If you want to add a new rule, you need to register the rule in `rules/provider.go` and update the `docs/rules/README.md`. Please refer to https://github.com/terraform-linters/tflint-ruleset-aws/pull/101

Also, You can use the generator to automate the work required to add new rules. See also here:
https://github.com/terraform-linters/tflint-ruleset-aws#add-a-new-rule

Thank you for your contribution :)

rules/aws_elasticache_replication_group_default_parameter_group.go
```diff
@@ -0,0 +1,65 @@
...
+func (r *AwsElastiCacheReplicationGroupDefaultParameterGroupRule) Severity() string {
+	return tflint.NOTICE
+}
+
+// Link returns the rule reference link
+func (r *AwsElastiCacheReplicationGroupDefaultParameterGroupRule) Link() string {
+	return project.ReferenceLink(r.Name())
+}
+
+var defaultElastiCacheParameterGroupRegexp = regexp.MustCompile("^default")
```
wata727, line 46: The `defaultElastiCacheParameterGroupRegexp` is already declared in here:
https://github.com/terraform-linters/tflint-ruleset-aws/blob/eaa469f7b0b1cddcdb34ae1fce6848e86990ce0e/rules/aws_elasticache_cluster_default_parameter_group.go#L46

Perhaps it's better to declare a variable name for this rule here.

Rihoj:
@wata727 I would like to also add in the rules for invalid parameter group, security group, and subnet group. However, I am not sure if I should try to add these into the API folder or not. I tried running `go run ./rules/api/generator` and received an error about build constraints. When I ran `go run ./rules/api/generator/main.go` it complained it couldn't find a file.

Are you wanting these rules inside the API folder and if so how would I go about creating these? If not am I good to create them with the same process as the rules I just added.

wata727:
The rules under the `rules/api` directory are a list of rules that are active only when `deep_check` is enabled, and if you want to create a rule like `aws_elasticache_cluster_invalid_parameter_group`, you need to put it under this directory.

These rules are automatically generated from definition files and you need to add a definition file like https://github.com/terraform-linters/tflint-ruleset-aws/blob/ef81b2ce272c396443598c6f37afa60b595f08c6/rules/api/definitions/aws_elasticache_cluster.hcl.

However, this is a bit complicated, so it seems good to get it out of the scope of this PR.

> I tried running go run ./rules/api/generator and received an error about build constraints.

Oops, this seems like a bug...

wata727:
👍

wata727:
> Oops, this seems like a bug...

Sorry, it is wrong. You can run this generator like the following:

```console
cd rules/api
go run -tags generators ./generator
```

rules/aws_elasticache_replication_group_invalid_type.go
```diff
@@ -0,0 +1,126 @@
...
+	nodeTypes     map[string]bool
+}
+
+// NewAwsElastiCacheReplicationGroupInvalidTypeRule returns new rule with default attributes
+func NewAwsElastiCacheReplicationGroupInvalidTypeRule() *AwsElastiCacheReplicationGroupInvalidTypeRule {
+	return &AwsElastiCacheReplicationGroupInvalidTypeRule{
+		resourceType:  "aws_elasticache_replication_group",
+		attributeName: "node_type",
+		nodeTypes: map[string]bool{
+			"cache.t2.micro":     true,
```
PatMyron, line 24: This list could probably be shared with the same list here:
https://github.com/terraform-linters/tflint-ruleset-aws/blob/c34979e4c88c750ff97b302ce56209bdc70f0b46/rules/aws_elasticache_cluster_invalid_type.go#L24

PatMyron: https://github.com/terraform-linters/tflint-ruleset-aws/pull/214

Issue #26, opened 2019-02-13
feat: detect iam policy documents greater or near 2048 maximum char limit


nitrocode:
AWS has a max char limit on policies of 2048 chars. This isn't found on terraform plan but could be checked in tflint.

What do you think?

PR #153, opened 2021-08-08
aws_iam_policy_length

Rihoj:
Added a rule to check against IAM policy length. fixes terraform-linters/tflint-ruleset-aws#26

I am not sure how you would like to have an extraordinarily long policy shown in the documentation and alerts so I did my best. If you would like a change just let me know how and I will get it updated.


rules/aws_iam_policy_invalid_length.go
```diff
@@ -0,0 +1,61 @@
...
+func (r *AwsIAMPolicyInvalidLengthRule) Check(runner tflint.Runner) error {
+	return runner.WalkResourceAttributes(r.resourceType, r.attributeName, func(attribute *hcl.Attribute) error {
+		var policy string
+		err := runner.EvaluateExpr(attribute.Expr, &policy, nil)
+
+		return runner.EnsureNoError(err, func() error {
+			if len(policy) > 2048  {
+				runner.EmitIssueOnExpr(
+					r,
+					fmt.Sprintf("The policy length is %d characters and is limited to 2048 characters.", len(policy)),
```
wata727, line 54: 2048 characters is a limit for inline policies. The managed policy limit is 6144 characters and should not count white spaces.
https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_iam-quotas.html#reference_iam-quotas-entity-length

rules/aws_iam_policy_invalid_length.go
```diff
@@ -0,0 +1,61 @@
...
+func NewAwsIAMPolicyInvalidLengthRule() *AwsIAMPolicyInvalidLengthRule {
+	return &AwsIAMPolicyInvalidLengthRule{
+		resourceType:  "aws_iam_policy",
+		attributeName: "policy",
+	}
+}
+
+// Name returns the rule name
+func (r *AwsIAMPolicyInvalidLengthRule) Name() string {
+	return "aws_iam_policy_invalid_length"
```
wata727, line 26: Personally, I prefer names like `aws_iam_policy_too_long_policy` for such rules. Naming such as `{resource}_invalid_{attribute}` is for automatic generation, and it is better to have an obvious name for the violation.

wata727 (commented):
This is a good start. Thanks!

I think we often use the [`aws_iam_policy_document` data source](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/iam_policy_document) to declare IAM policies, and I couldn't think of a way to support the case, so I've put this issue on hold. However, it's a good idea to support just the string literal pattern.

rules/aws_iam_policy_invalid_length.go
Rihoj: Ahh. Yeah, that makes sense. Given it's entity based, and we would then also have to look at what it's being attached to we may not be able to apply a strict check to this except the highest option which is actually 10,240 (inline role policy). 

However, we could look at doing this for specifically as
aws_iam_group_policy: 5120
aws_iam_role_policy: 10240
aws_iam_user_policy: 2048

Unless, you know a way we can find what entity the policy is going to be attached to I can't think of any other option.

rules/aws_iam_policy_invalid_length.go
Rihoj: I will keep that in mind, and depending on where we go with this I will update it accordingly.

Rihoj:
As far as the data source goes, I wasn't even aware that one existed as I don't commonly write policies as code yet. (not enough buy in unfortunately.)

rules/aws_iam_policy_invalid_length.go
wata727: Hmm, since `aws_iam_policy` always creates a [customer managed policy](https://docs.aws.amazon.com/IAM/latest/UserGuide/access_policies_managed-vs-inline.html#customer-managed-policies), isn't the upper limit 6,144?
I think the upper limit for `inline_policy` blocks such as `aws_iam_role` resources will be 10,240.
https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/iam_role#example-of-exclusive-inline-policies

Rihoj: @wata727 Hmm. yeah, That makes more sense... I think the wording in this document kept throwing me off: https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_grammar.html#policies-grammar-notes

Reading the limits page again. I think you are right aws_iam_policy should be good to set to 6144. And I am not sure the inlines can be checked given the docs " But the total aggregate policy size (the sum size of all inline policies) per entity cannot exceed the following quotas"

My apologies on the confusion here.

Rihoj:
Depending on how this one goes I may pick up the inline ones next.

wata727:
👍

Issue #149, opened 2021-08-03
rule: aws_iam_policy_document - Sid string check


andormarkus:
Hi Community,

The plugin currently does not validates [`aws_iam_policy_document`]() `sid_string`.
Based on [AWS documentation](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_grammar.html#policies-grammar-notes) basic alphanumeric characters (A-Z,a-z,0-9) are the only allowed characters in the Sid value.

It would be great if there would be rule to check the sid_string.

Thanks,
Andor

wata727:
#155 was a rule for the `aws_iam_policy` resource. This issue is a rule request for the `aws_iam_policy_document` data resource, so reopen the issue.

gilbert-hsu:
Should we valid sid pattern even sid absent in statement ?

```
Error: The policy's sid ("") does not match "^[a-zA-Z0-9]+$". 
(aws_iam_policy_sid_invalid_characters)
  on iam-policy.tf line 22:
  22:   policy = <<EOF
  23: {
  24:   "Version": "2012-10-17",
  25:   "Statement": [
  26:     {
  27:       "Action": [],
  34:       "Effect": "Allow",
  35:       "Resource": "*"
  36:     }
  37:   ]
  38: }
  39: EOF
  ```

**Comment by wata727** [comment]
@gilbert-hsu Oops, this is a false positive. We have released v0.7.1 with this fix, so please try it.

**Comment by gilbert-hsu** [comment]
Thanks a lot！
v0.7.1 is fine in my environment！

**Comment by kyblik** [comment]
Got a failing use case even though this is a valid policy (notice the lack of `[]` in `Statement`):

```
resource "aws_iam_policy" "failing" {
  name   = "failing"
  policy = <<EOF
{
  "Version": "2012-10-17",
  "Statement": {
    "Effect": "Allow",
    "Action": "sts:AssumeRole",
    "Resource": "*"
  }
}
EOF
}
``` 

```
resource "aws_iam_policy" "passing" {
  name   = "passing"
  policy = <<EOF
{
  "Version": "2012-10-17",
  "Statement": [{
    "Effect": "Allow",
    "Action": "sts:AssumeRole",
    "Resource": "*"
  }]
}
EOF
}
``` 

Error messages:

```
Failed to check ruleset. An error occurred:

[...]

PR #155, opened 2021-08-09
aws_iam_policy_sid

Rihoj:
add invalid sid rules fixes terraform-linters/tflint-ruleset-aws#149

This should loop through all SIDs in a policy and make sure all of them fit the pattern.

### DIFF
--- a/rules/aws_iam_policy_sid_invalid_characters.go
+++ b/rules/aws_iam_policy_sid_invalid_characters.go
@@ -6,7 +6,14 @@ import (
 	"github.com/terraform-linters/tflint-plugin-sdk/tflint"
 	"github.com/terraform-linters/tflint-ruleset-aws/project"
 	"regexp"
+	"fmt"
 )
+type AwsIAMPolicySidInvalidCharactersStatementStruct struct {
+	Sid string `json:"Sid"`
+}
+type AwsIAMPolicySidInvalidCharactersPolicyStruct struct {
+	Statement []AwsIAMPolicySidInvalidCharactersStatementStruct `json:"Statement"`
+}
 
 // AwsIAMPolicySidInvalidCharactersRule checks for invalid characters in SID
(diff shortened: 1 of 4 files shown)


rules/aws_iam_policy_sid_invalid_characters.go
```diff
@@ -0,0 +1,74 @@
...
+	return runner.WalkResourceAttributes(r.resourceType, r.attributeName, func(attribute *hcl.Attribute) error {
+		var val string
+		err := runner.EvaluateExpr(attribute.Expr, &val, nil)
+		var unMarshaledPolicy interface{}
+		err = json.Unmarshal([]byte(val), &unMarshaledPolicy)
+		if err != nil {
+			return err
+		}
+		policy := unMarshaledPolicy.(map[string]interface{})
+		statements := policy["Statement"].([]interface{})
```
wata727, line 58: Casting `interface{}` can cause panic. In this case, it's a good idea to declare a structure for the policy and unmarshal it there.
https://gobyexample.com/json

rules/aws_iam_policy_sid_invalid_characters.go
```diff
@@ -0,0 +1,74 @@
...
+
+		return runner.EnsureNoError(err, func() error {
+			for _, e := range statements {
+				statement := e.(map[string]interface{})
+				if r.validCharacters.MatchString(statement["Sid"].(string)) == false {
+					runner.EmitIssueOnExpr(
+						r,
+						"The policy's sid contains invalid characters.",
```
wata727, line 66: The error message should clearly indicate why it is invalid. Perhaps it would be nice to include a regexp pattern.

Rihoj:
@wata727 those should be fixed. Thank you for the feedback.

wata727:
Thanks!

Issue #33, opened 2020-01-22
tflint deep IAM policy


ian-bartholomew:
Hello, we want to run this in CI, and be able to also run the deep inspection, but its not clear what kind of IAM permissions are needed.  We want to run this in CI, and grant it the least privileged IAM policy.

Thanks!

rquadling:
Hi.

Please can we have a complete list of required permissions for tflint.

We have a specific user that is used to plan and apply using terraform and all is working there.

We've introduced tflint into our process and we now requiring additional access...
```
Failed to check ruleset. An error occurred:

Error: Failed to check `aws_instance_invalid_key_name` rule: An error occurred while invoking DescribeKeyPairs; UnauthorizedOperation: You are not authorized to perform this operation.
	status code: 403, request id: *********
```

amongst other things.

As it stands, the user used to deploy the terraform plan does NOT require these permissions and so we would not have granted that.

If this is something that is available in code (but not documented), then potentially the list of calls being made by the ruleset could be extracted and the logged. That would certainly be a start?

Some sort of auto-discovery?

wata727:
Ah, that makes sense. It's easy to document the list of required permissions because just check the implementation under the [`api`](https://github.com/terraform-linters/tflint-ruleset-aws/tree/8047cf5d9d74e1175e1fb2481853577f5ca6dd1e/rules/api) directory.

I haven't been able to work on this for other work, but I'd be happy to review it if you open a PR.

PR #157, opened 2021-08-15
docs: Add required IAM policy document for deep checking

wata727:
Fixes https://github.com/terraform-linters/tflint-ruleset-aws/issues/33

### DIFF
--- a/docs/deep_checking.md
+++ b/docs/deep_checking.md
@@ -109,3 +109,37 @@ This plugin fetches credentials in the same way as Terraform. See [this document
 ### Assume role
 
 This plugin can assume a role in the same way as Terraform. See [this documentation](https://registry.terraform.io/providers/hashicorp/aws/latest/docs#assume-role).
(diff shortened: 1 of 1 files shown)


PatMyron:
Wonder if this policy could be automatically generated by scraping `AwsClient` calls in [`rules/api/`](https://github.com/terraform-linters/tflint-ruleset-aws/tree/master/rules/api)

In the meantime, could monitor new `AwsClient` calls with sourcegraph monitoring to at least get manually alerted:

https://sourcegraph.com/search?q=context:global+AwsClient.+r:^github\.com/terraform-linters/tflint-ruleset-aws$+-file:_test\.go$+type:diff&patternType=literal

wata727:
Looks good. To be honest, deep checking features are very minimal now.
I think there is a problem that it will be slower if there are more rules, but I believe that parallel inspection or use with language server can improve the experience.

PR #159, opened 2021-08-15
Add aws_iam_group_policy too long rule

Rihoj:
I went ahead and namespaced the random function in the other class. I wonder if there would be a good way to save this in a "helper" location to be pulled in instead.


rules/aws_iam_group_policy_too_long.go
```diff
@@ -0,0 +1,65 @@
...
+func (r *AwsIAMGroupPolicyTooLongRule) Severity() string {
+	return tflint.ERROR
+}
+
+// Link returns the rule reference link
+func (r *AwsIAMGroupPolicyTooLongRule) Link() string {
+	return project.ReferenceLink(r.Name())
+}
+
+// TODO: Write the details of the inspection
```
wata727, line 46: Please remove the TODO comments 🚮

rules/aws_iam_group_policy_too_long.go
```diff
@@ -0,0 +1,65 @@
...
+// Check checks the length of the policy
+func (r *AwsIAMGroupPolicyTooLongRule) Check(runner tflint.Runner) error {
+	return runner.WalkResourceAttributes(r.resourceType, r.attributeName, func(attribute *hcl.Attribute) error {
+		var policy string
+		err := runner.EvaluateExpr(attribute.Expr, &policy, nil)
+		whitespaceRegex := regexp.MustCompile(`\s+`)
```
wata727, line 52: Go's regexp compile is slow, so I prefer to compile the regexp in `NewAwsIAMGroupPolicyTooLongRule` rather than the `Check` function, which runs over and over again.
https://github.com/terraform-linters/tflint-ruleset-aws/blob/20a14646ec7402659b997ee70858e59968af25cf/rules/models/aws_acm_certificate_invalid_certificate_body.go#L30

The good point about this way is that if you have a problem with the regexp, you will get a panic early and notice the problem (`NewAwsIAMGroupPolicyTooLongRule` will be always called).

docs/rules/aws_iam_group_policy_too_long.md
```diff
@@ -0,0 +1,34 @@
...
+1 issue(s) found:
+Error: The policy length is %d characters and is limited to 5120 characters. (aws_iam_policy_too_long_policy)
+  on template.tf line 6:
+   6:   policy = "{POLICY}" // Policy is too long,!
+
+```
+
+## Why
+
+Terraform does not check against this rule, but it will error if an apply is attempted.
```
wata727, line 30: It would be nicer to have a link to the underlying documentation:
https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_iam-quotas.html#reference_iam-quotas-entity-length

wata727:
> I wonder if there would be a good way to save this in a "helper" location to be pulled in instead.

Looks good

Rihoj:
@wata727 sorry for the delay, life has been busy. I hope this meets your expectations.

wata727:
Thanks!

PR #170, opened 2021-08-28
Update valid regions for the aws_s3_bucket_invalid_region rule

wata727:
Added the following regions:

- `eu-south-1`
- `af-south-1`
- `us-gov-east-1`
- `us-gov-west-1`


PatMyron:
think this could be pulled from the AWS SDK models along with the other mappings

PR #176, opened 2021-08-31
Fix false positive for IAM policy document without Sid

wata727:
See https://github.com/terraform-linters/tflint-ruleset-aws/issues/149#issuecomment-908876081

Probably, this is a valid case :)


Rihoj (approved):
This all makes sense to me.

PR #211, opened 2021-12-08
Add missing t4g ElastiCache node types

acastro2:
Missing t4g instances listed: https://docs.aws.amazon.com/AmazonElastiCache/latest/red-ug/CacheNodes.SupportedTypes.html


wata727 (approved):
Good catch!

PR #214, opened 2021-12-09
Extract ElastiCache node types to utils

wata727:
Follow up of https://github.com/terraform-linters/tflint-ruleset-aws/pull/211 https://github.com/terraform-linters/tflint-ruleset-aws/pull/213

This PR extracts valid/previous ElastiCache node types to utils from each rule definition. This provides consistent inspection in clusters and replications.


rules/utils.go
```diff
@@ -0,0 +1,97 @@
...
+	"cache.m3.large":   true,
+	"cache.m3.xlarge":  true,
+	"cache.m3.2xlarge": true,
+	"cache.r3.large":   true,
+	"cache.r3.xlarge":  true,
+	"cache.r3.2xlarge": true,
+	"cache.r3.4xlarge": true,
+	"cache.r3.8xlarge": true,
+	"cache.c1.xlarge":  true,
+	"cache.t1.micro":   true,
```
PatMyron, line 96: thinking prefixes alone could identify previous generation types:
```
	"cache.m1": true,
	"cache.m2": true,
	"cache.m3": true,
	"cache.r3": true,
	"cache.c1": true,
	"cache.t1": true,
```

PR #217, opened 2021-12-21
ap-southeast-3 (Jakarta)

PatMyron:
prev: https://github.com/terraform-linters/tflint-ruleset-aws/pull/170, https://github.com/terraform-linters/tflint/pull/442
https://aws.amazon.com/blogs/aws/now-open-aws-asia-pacific-jakarta-region/

---

wondering what the lag will be to pull this from the AWS SDK models along with the other mappings?
https://github.com/terraform-linters/tflint-ruleset-aws/blob/7ef3658eb5870c26136285ed74923d12de960549/rules/models/mappings/s3.hcl#L24


rules/models/aws_route53_zone_association_invalid_vpc_region.go
```diff
@@ -51,6 +52,7 @@ func NewAwsRoute53ZoneAssociationInvalidVpcRegionRule() *AwsRoute53ZoneAssociati
 			"sa-east-1",
 			"ca-central-1",
 			"cn-north-1",
+			"cn-northwest-1",
```
PatMyron, line 55: this also seemed missing, so I added it along with this change

bendrucker:
Added in #216

PatMyron:
@bendrucker `rules/models/*.go` enums pick this up from `go generate` (removed those edits in the second commit), but this S3 enum seems still manually maintained in `rules/` as far as I can tell

PR #290, opened 2022-01-07
automating maintenance with Github actions

PatMyron:
similar to https://github.com/aws-cloudformation/cfn-lint/pull/1792

currently running and updating this PR in my fork: https://github.com/PatMyron/tflint-ruleset-aws/pull/2


.github/workflows/maintenance.yaml
```diff
@@ -0,0 +1,29 @@
+on:
+  push:
+    branches: [ master ]
+  schedule:
+    - cron: '*/5 * * * *'
```
wata727, line 5: It seems aggressive to run every 5 minutes. I thought it was enough to run it daily. What do you think?

.github/workflows/maintenance.yaml
```diff
@@ -0,0 +1,29 @@
...
+      - uses: actions/setup-go@v2
+        with:
+          go-version: 1.17
+      - run: |
+          go get github.com/aws/aws-sdk-go
+          go mod tidy
+          go generate ./...
+          cd tools/provider-schema
+          terraform init -upgrade
+          terraform providers schema -json > schema.json
```
wata727, line 23: Ah, I'm surprised that ubuntu-latest includes Terraform by default.
https://github.com/actions/virtual-environments/blob/cd453927cac82e4dc2658bb65d42b2c97eb83212/images/linux/scripts/installers/terraform.sh

However, this action may fail when a new version is released. It seems good to pin the version using setup-terraform.
https://github.com/hashicorp/setup-terraform

PatMyron: Thanks for the pointer, considered installing `tfenv` as [recommended in our docs](https://github.com/terraform-linters/tflint-ruleset-aws/tree/master/tools/provider-schema) but knew there were issues installing it when terraform was already installed. I'll look into that instead

.github/workflows/maintenance.yaml
```diff
@@ -0,0 +1,29 @@
...
+      - uses: peter-evans/create-pull-request@v3
```
bendrucker, line 24: Answering my own question here, this will continuously update the PR

https://github.com/peter-evans/create-pull-request#action-behaviour

.github/workflows/maintenance.yaml
```diff
@@ -0,0 +1,29 @@
...
+        with:
+          commit-message: |
+            autogenerated maintenance
+          title: autogenerated maintenance
```
bendrucker, line 28:
```suggestion
          title: Generate rules from terraform-provider-aws
```

Would be nice to append `@version` as well, but that may increase the difficulty of targeting the PR. Dependabot/Renovate do something like this but probably w/ considerable logic involved.

.github/workflows/maintenance.yaml
```diff
@@ -0,0 +1,29 @@
...
+          go get github.com/aws/aws-sdk-go
+          go mod tidy
+          go generate ./...
+          cd tools/provider-schema
+          terraform init -upgrade
+          terraform providers schema -json > schema.json
+      - uses: peter-evans/create-pull-request@v3
+        with:
+          commit-message: |
+            autogenerated maintenance
```
bendrucker, line 27:
```suggestion
            Generate rules from terraform-provider-aws. Automatically generated via the [maintenance](.github/workflows/maintenance.yaml) workflow.
```

.github/workflows/maintenance.yaml
bendrucker, line 24: When this pull request is created, it will not trigger workflows, since GH Actions forbids recursion when using `secrets.GITHUB_TOKEN`.

bendrucker (approved):
Nice, thanks for this!

bendrucker (commented):
We do need to figure out the recursive check runs. This is why I haven't undertaken more auto-generation already. Putting a PAT in is a fair bit of inconvenience but we can do it at the org level I guess.

.github/workflows/maintenance.yaml
PatMyron: not elegant, but I usually just close/re-open the PRs to manually kick off the tests:
https://github.com/peter-evans/create-pull-request/issues/48

.github/workflows/maintenance.yaml
PatMyron: also not elegant, but I've just manually appended version numbers to autogenerated PR titles in the past due to complexities of figuring it out automatically:
https://github.com/aws-cloudformation/cfn-lint-visual-studio-code/pull/76

.github/workflows/maintenance.yaml
```diff
@@ -0,0 +1,32 @@
...
+          submodules: true
+      - uses: actions/setup-go@v2
+        with:
+          go-version: 1.17
+      - run: |
+          go get github.com/aws/aws-sdk-go
+          go mod tidy
+          go generate ./...
+          cd tools/provider-schema
+          terraform init -upgrade
```
PatMyron, line 22: @wata727 @bendrucker anyone have any ideas why `terraform init -upgrade` doesn't seem to be upgrading the AWS provider here? https://github.com/patmyron/tflint-ruleset-aws/pull/2

That command upgrades the AWS provider successfully for me in other environments I've tried it...

PatMyron:
Merging as-is to automatically pick up new AWS types to continue mapping
Feel free to continue tweaking / iterating with any improvements if desired

.github/workflows/maintenance.yaml
PatMyron: usually unsubscribe from notifications from the generated PRs

PR #311, opened 2022-03-01
Don't use inline rule blocks within aws_security_group (terraform-linters#304)

vikram2784:


wata727:
Thank you for working on this. But I'm wondering if this rule is right.
Rule conflicts are an issue to avoid, but I think it is good to be able to represent a security group by a single resource. I think this is the reason why the AWS provider allows inline rules despite conflict issues.

There are two options we can take:

1. Allows the user to choose between inline style and resource style and turns the default off.
2. Warn only if there is a conflict for the same security group.

1 is not a bad option, but I personally prefer 2.

vikram2784:
@wata727 Thanks for the review. Yes, that was the reason I marked this rule as warning instead of error.  There are few discussions on some forums that talk about this issue. However, I agree with your perspective.  Determining a conflict will need us to build a context/state

I was looking at implementing Point 2, but couldn't figure how to parse the value of `security_group_id` in `aws_security_group_rule` . We will need to extract the value of security_group NAME from `aws_security_group.<NAME>.id` (expression). Note -- This won't work for literal values (like "sg-12345")

wata727:
> We will need to extract the NAME from security_group_id's value i.e aws_security_group.<NAME>.id (from an expression). Note -- This won't work for literal values (like "sg-12345")

Yes, I think this is a good approach.  As you say, It doesn't work for literals, but that's acceptable as a limitation of this rule.

wata727:
@vikram2784 Any updates?

vikram2784:
@wata727 Sorry, I havent had the time to look into this. Will look into it next couple of weeks

wata727:
This PR has been closed due to lack of recent activity from the author. Feel free to reopen if you're ready.

PR #315, opened 2022-03-17
aws_elastic_beanstalk_environment_invalid_name_format

samhpickering:
Elastic Beanstalk environment names must match the regex ^[a-zA-Z0-9][a-zA-Z0-9-]+[a-zA-Z0-9]$ or will create an error when being provisioned.
There is already a rule aws_elastic_beanstalk_environment_invalid_name generated from the SDK, but that only includes the length constraint of 4 to 40 characters.


wata727:
👍 

Finally, please add the documentation to [docs/rules/README.md](https://github.com/terraform-linters/tflint-ruleset-aws/blob/58eef995422803bfdf0b65e2240a790a008d1a07/docs/rules/README.md.tmpl)

wata727:
We've made a major change to the upstream, so I've resolved the conflict and pushed it to your repository.
Merge this pull request. Thank you for your contribution.

Issue #331, opened 2022-04-04
Support Provider aliases


suzuki-shunsuke:
## Problem

Terraform supports defining multiple configurations for the same provider, and select which one to use on a per-resource or per-module basis.

https://www.terraform.io/language/providers/configuration#alias-multiple-provider-configurations

But tflint doesn't support this, so even if multiple configurations are defined only one configuration is used.
This raises a problem in Deep Checking.

## How to reproduce

```console
$ tflint -v
TFLint version 0.35.0
+ ruleset.aws (0.13.2)
```

.tflint.hcl

```hcl
plugin "aws" {
  enabled = true
  version = "0.13.2"
  source  = "github.com/terraform-linters/tflint-ruleset-aws"

  deep_check = true
}
```

main.tf

```tf
terraform {
  required_version = ">= 0.15"
  required_providers {
    aws = {
      source  = "hashicorp/aws"
      version = "4.0.0"
    }
  }
}

provider "aws" {
  region = "ap-northeast-1"
}

provider "aws" {
  region = "us-east-1"
  alias  = "us-east-1"
}

resource "aws_instance" "reverse_proxy_a01" {
  ami           = "ami-088da9557aae42f39"
  instance_type = "t3.micro"
  key_name      = "bootstrap"
}
```

```console
$ tflint
2 issue(s) found:

Error: "ami-088da9557aae42f39" is invalid AMI ID. (aws_instance_invalid_ami)

  on main.tf line 21:
  21:   ami           = "ami-088da9557aae42f39"

Error: "bootstrap" is invalid key name. (aws_instance_invalid_key_name)

  on main.tf line 23:
  23:   key_name      = "bootstrap"
```

This error occurs even if Key Pair and AMI exist in the region `ap-northeast-1`, because tflint gets resources from `us-east-1` using the alias `us-east-1`.

## Solution

* Change `Runner.AwsClient` to map `map[string]*Client` and creates clients per Provider Configuration
* Check resource's `provider` attribute and get the appropriate Client from map of clients

I tried to implement this, but it doesn't work well yet.

https://github.com/terraform-linters/tflint-ruleset-aws/pull/332

I'll appreciate if you help me.
It is difficult to debug because I don't know how to output the plugin log.

## Reference

* https://www.terraform.io/language/providers/configuration#alias-multiple-provider-configurations

suzuki-shunsuke:
What's the problem?

[...]

PR #332, opened 2022-04-04
[WIP] feat: support provider aliases

suzuki-shunsuke:
https://github.com/terraform-linters/tflint-ruleset-aws/issues/331

This is still work in progress, and it doesn't work well.

### DIFF
--- a/aws/runner.go
+++ b/aws/runner.go
@@ -3,6 +3,7 @@ package aws
 import (
 	"fmt"
 	"log"
+	"strings"
 
 	"github.com/hashicorp/hcl/v2"
 	"github.com/terraform-linters/tflint-plugin-sdk/hclext"
@@ -48,6 +49,7 @@ func (r *Runner) AwsClient(attributes hclext.Attributes) (*Client, error) {
 		if err := r.EnsureNoError(r.EvaluateExpr(attr.Expr, &provider, nil), func() error { return nil }); err != nil {
 			return nil, err
 		}
+		provider = strings.TrimPrefix(provider, "aws.")
 	}
 
 	awsClient, ok := r.AwsClients[provider]
--- a/aws/runner.go
+++ b/aws/runner.go
@@ -1,9 +1,11 @@
 package aws
 
 import (
+	"fmt"
 	"log"
 
 	"github.com/hashicorp/hcl/v2"
+	"github.com/terraform-linters/tflint-plugin-sdk/hclext"
 	"github.com/terraform-linters/tflint-plugin-sdk/tflint"
 )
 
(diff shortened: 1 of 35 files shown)


suzuki-shunsuke:
This seems not to work well.
https://github.com/suzuki-shunsuke/tflint-ruleset-aws/blob/f1695c42740c3d106e9b9b2e37a5ea5f39b3dd32/aws/runner.go#L47-L51

suzuki-shunsuke:
What's the problem?

https://github.com/terraform-linters/tflint-ruleset-aws/runs/5814975813?check_suite_focus=true

```
panic: open aws-sdk-go/models/apis/account/2021-02-01/api-2.json: no such file or directory

goroutine 1 [running]:
main.main()
	/home/runner/work/tflint-ruleset-aws/tflint-ruleset-aws/rules/models/generator/main.go:64 +0xb71
exit status 2
rules/models/models.go:3: running "go": exit status 1
Error: Process completed with exit code 1.
```

aws/runner.go
```diff
@@ -1,41 +1,63 @@
...
 		PluginConfig: config,
-		AwsClient:    client,
+		AwsClients:   clients,
 	}, nil
 }
 
+func (r *Runner) AwsClient(attributes hclext.Attributes) (*Client, error) {
+	provider := "aws"
+	if attr, exists := attributes["provider"]; exists {
+		if err := r.EnsureNoError(r.EvaluateExpr(attr.Expr, &provider, nil), func() error { return nil }); err != nil {
```
suzuki-shunsuke, line 48: It doesn't work well.

suzuki-shunsuke: You have to get the alias from the attribute.

e.g.

```tf
resource "aws_instance" "foo" {
  provider = aws.west # You have to get the alias "west".

  # ...
}
```

suzuki-shunsuke: To support Terraform Modules, probably you have to fix something, but unfortunately I don't know well.

```tf
module "aws_vpc" {
  source = "./aws_vpc"
  providers = {
    aws = aws.west
  }
}
```

wata727: > To support Terraform Modules, probably you have to fix something, but unfortunately I don't know well.

Umm, it's a hard problem. To solve this problem, we need to refer to the provider map of the module call.

The `Check()` method of each rule is called for each module (The root module is also one of the modules). So, `GetModuleContent` basically cannot access the parent module containing a module call that calls itself.

The easiest way is to stop support for alias in child modules. If you haven't faced this issue with aliases in modules, you'll be able to accept it.

The best way is to add an API that references the parent module. `GetModuleContent` has options for the module context, imagine something like `ParentModuleCtxType`.
https://github.com/terraform-linters/tflint-plugin-sdk/blob/v0.10.1/tflint/option.go#L18-L19

You should probably be able to achieve the intended behavior by getting the provider map with `Check()` method in the `RuleSet` and then caching it in the Runner.
https://github.com/terraform-linters/tflint-ruleset-aws/blob/7523ed68628e673e2e76572e32634fb5002b5285/aws/ruleset.go#L46-L59

suzuki-shunsuke:
```tf
  provider      = aws.us-east-1
```

```
07:23:43 runner_eval.go:40: [INFO] unevaluable expression found in main.tf:27. TFLint ignores unevaluable expressions.
07:23:43 [ERROR] interceptor/logging.go:18: failed to gRPC request: direction=plugin2host method=/proto.Runner/EvaluateExpr err="rpc error: code = FailedPrecondition desc = unevaluable expression found in main.tf:27"
```

Hmm... It fails to parse the provider attribute.

suzuki-shunsuke:
> > What's the problem?
> > https://github.com/terraform-linters/tflint-ruleset-aws/runs/5814975813?check_suite_focus=true
> 
> It seems to be caused by an update of the aws-sdk-go module. Some rules are generated from the API definition files included in aws-sdk, but the file to be referenced may have changed due to the update.
> 
> If you don't need to update the aws-sdk-go module for this change, I recommend that you revert it or open it as another pull request.

I see. Thank you for your explanation.
`rules/models/aws-sdk-go` was updated automatically, so I committed it.
I reverted. faa5f65 d6e45af

aws/runner.go
wata727: `EvaluateExpr` limits the types of expressions that can be evaluated, so cannot evaluate expressions like `aws.*`.
https://github.com/terraform-linters/tflint/blob/v0.35.0/tflint/runner_eval.go#L93-L104

In the Terraform core, it seems that alias doesn't support evaluation, just get it with `gohcl.DecodeExpression`. The same way can be used.
https://github.com/hashicorp/terraform/blob/v1.1.7/internal/configs/provider.go#L58

suzuki-shunsuke: Thank you for your help.

In the Terraform core, resource's "provider" attribute seems to be parsed here.
https://github.com/hashicorp/terraform/blob/3fbedf25430ead97eb42575d344427db3c32d524/internal/configs/resource.go#L136-L140

suzuki-shunsuke:
It is hard to resolve the conflict, so maybe I'll create a new pull request.

suzuki-shunsuke:
https://github.com/terraform-linters/tflint-ruleset-aws/pull/342

PR #333, opened 2022-04-04
style: format rules/api/rule.go.tmpl and run `go generate ./...`

suzuki-shunsuke:
Format `format rules/api/rule.go.tmpl` and run `go generate ./...`.

```console
$ go generate ./...
```

Fix the indent.


wata727 (approved):
👍

PR #342, opened 2022-05-05
feat: support provider aliases

suzuki-shunsuke:
https://github.com/terraform-linters/tflint-ruleset-aws/issues/331

Follow up https://github.com/terraform-linters/tflint-ruleset-aws/pull/332#issuecomment-1118514083

### DIFF
--- a/aws/runner.go
+++ b/aws/runner.go
@@ -33,6 +33,13 @@ func NewRunner(runner tflint.Runner, config *Config) (*Runner, error) {
 			}
 			clients[k] = client
 		}
+		if _, ok := credentials["aws"]; !ok {
+			client, err := NewClient(config.toCredentials())
+			if err != nil {
+				return nil, err
+			}
+			clients["aws"] = client
+		}
 	}
 
 	return &Runner{
--- a/aws/runner.go
+++ b/aws/runner.go
@@ -2,7 +2,6 @@ package aws
 
 import (
 	"fmt"
-	"strings"
 
 	"github.com/hashicorp/hcl/v2"
 	"github.com/terraform-linters/tflint-plugin-sdk/hclext"
@@ -46,10 +45,16 @@ func NewRunner(runner tflint.Runner, config *Config) (*Runner, error) {
 func (r *Runner) AwsClient(attributes hclext.Attributes) (*Client, error) {
 	provider := "aws"
 	if attr, exists := attributes["provider"]; exists {
-		if err := r.EnsureNoError(r.EvaluateExpr(attr.Expr, &provider, nil), func() error { return nil }); err != nil {
-			return nil, err
+		providerConfigRef, diags := decodeProviderConfigRef(attr.Expr, "provider")
+		if diags.HasErrors() {
+			logger.Error("parse resource provider attribute: %s", diags)
+			return nil, diags
 		}
-		provider = strings.TrimPrefix(provider, "aws.")
+		if providerConfigRef.Alias != "" {
+			provider = providerConfigRef.Alias
+		} else {
+			provider = providerConfigRef.Name
+		}
 	}
 
 	awsClient, ok := r.AwsClients[provider]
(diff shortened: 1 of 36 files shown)


wata727:
Umm... As you say, it seems that it is necessary to reimplement the `decodeProviderConfigRef`.
https://github.com/hashicorp/terraform/blob/3fbedf25430ead97eb42575d344427db3c32d524/internal/configs/resource.go#L498

I think we doesn't need to reproduce all implementations completely, for example, support for legacy configuration language can be ignored. The important thing is to get the list of traversals with `hcl.AbsTraversalForExpr` and get the root name and step name.

suzuki-shunsuke:
c3caa6c Copied required code.
It seems to work well.
I guess we have to take the original license into account.

suzuki-shunsuke:
Mozilla Public License, version 2.0
https://github.com/hashicorp/terraform/blob/main/LICENSE

aws/runner.go
```diff
@@ -10,31 +13,58 @@ import (
...
-
-		client, err = NewClient(config.toCredentials().Merge(credentials))
-		if err != nil {
-			return nil, err
+		for k, cred := range credentials {
+			client, err := NewClient(config.toCredentials().Merge(cred))
+			if err != nil {
+				return nil, err
+			}
+			clients[k] = client
```
wata727, line 34: I noticed that the client is not initialized as before if you don't declare any provider blocks. Previously, if you wrote a credential in `.tflint.hcl`, you should have been able to enable deep checking without a provider block. e.g.

```hcl
plugin "aws" {
  enabled = true
  access_key = "AWS_ACCESS_KEY",
  secret_key = "AWS_SECRET_KEY",
}
```

```hcl
resource "aws_instance" "foo" {
  ami = "invalid"
}
```

rules/api/aws_instance_invalid_ami.go
```diff
@@ -72,13 +72,18 @@ func (r *AwsInstanceInvalidAMIRule) Check(rr tflint.Runner) error {
 			continue
 		}
 
+		awsClient, err := runner.AwsClient(resource.Body.Attributes)
```
wata727, line 75: You need to get the provider attribute explicitly like any other rule.

rules/api/aws_launch_configuration_invalid_image_id.go
```diff
@@ -72,13 +72,18 @@ func (r *AwsLaunchConfigurationInvalidImageIDRule) Check(rr tflint.Runner) error
 			continue
 		}
 
+		awsClient, err := runner.AwsClient(resource.Body.Attributes)
```
wata727, line 75: Same as aws_instance_invalid_rule.

wata727:
> I guess we have to take the original code's license into account.

Yes. We changed TFLint's license 4 years ago to follow this case. MPL 2.0 is a weak copyleft and the Large Work is allowed under the same license.
https://github.com/terraform-linters/tflint/pull/245

rules/api/aws_launch_configuration_invalid_image_id.go
suzuki-shunsuke: Fixed. 821423f

rules/api/aws_instance_invalid_ami.go
suzuki-shunsuke: Fixed. 821423f

aws/runner.go
suzuki-shunsuke: f73855a2e4d5ab667d848e10eaa0c559f47febfa

suzuki-shunsuke: https://github.com/terraform-linters/tflint-ruleset-aws/blob/master/docs/deep_checking.md#shared-credentials

> If these configurations are defined in the provider block, they will also be taken into account. But the priority is lower than the above way.

Hmm... Maybe we have to fix f73855a according to the priority.

suzuki-shunsuke: The priority of `plugin` setting is higher than `provider`.

suzuki-shunsuke: Fixed. 313f0da9624d402e23056bdba9b990fe202e1cd8

wata727 (approved):
🎉

suzuki-shunsuke:
Thank you for your support!

PR #355, opened 2022-07-10
Add aws_security_group_rule_invalid_protocol rule

x-color:
Add `aws_security_group_rule_invalid_protocol` rule.

This rule checks whether `protocol` has an invalid value as protocol.

PR #356, opened 2022-07-10
Add aws_security_group_invalid_protocol rule

x-color:
Add `aws_security_group_invalid_protocol` rule.

This rule checks whether `protocol` in an inline rule block has an invalid value as protocol.


wata727 (commented):
There are same suggestions as https://github.com/terraform-linters/tflint-ruleset-aws/pull/355#pullrequestreview-1033605754

x-color:
@wata727 
Thank you for reviewing my PR!
I fixed them.

wata727:
👍

PR #362, opened 2022-08-12
Lambda runtime deprecation updates

PatMyron:
https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html#runtime-support-policy


wata727 (approved):
👍

PR #367, opened 2022-08-31
remove hardcoded S3 region rule

PatMyron:
prev: https://github.com/terraform-linters/tflint-ruleset-aws/pull/217, https://github.com/terraform-linters/tflint-ruleset-aws/pull/170, https://github.com/terraform-linters/tflint/pull/442

---

new region launch seems like a good time to stop manually hardcoding this list of regions:
https://aws.amazon.com/blogs/aws/now-open-aws-region-in-the-united-arab-emirates-uae/


wata727 (commented):
Thank you for working on this.

I agree that hard-coded lists are hard to maintain, but I believe such maintenance will be necessary to make this ruleset more valuable. I would like to maintain this list as necessary in the future, but what do you think?

PatMyron:
@wata727 since there are currently ~1400 SDK-based rules and only a few dozen manual rules, I don't think it's worth actively maintaining this particular manual rule a couple times a year for every new AWS region. Pruning a couple rules like this can save us a lot of maintenance while the value of the ruleset stays roughly the same since it was just one of ~1500 rules

wata727:
Correct. But if we remove this rule for that reason, I think we will not be able to add manually-maintained rules.
I believe that lists like regions can be maintained by the community. I would like the ruleset to be a place where users bring value.

PatMyron:
> we will not be able to add manually-maintained rules

The churn is specific to this rule because there will be dozens of new AWS regions over time. None of the other manually maintained enums have nearly as many values, so the other manually maintained rules are fine in my opinion

> maintained by the community

My concern is this rule would cause noise in the meantime until someone in the community steps up every time there's a new AWS region

wata727:
> My concern is this rule would cause noise in the meantime until someone in the community steps up every time there's a new AWS region

Agreed, but I believe that not many people will be affected by this noise. It's not too late to remove this rule, even after many users reported it was causing inconvenience.

PR #381, opened 2022-09-28
Adding lint testing for heredocs

amvapor:
Adding heredoc scanning, allowing someone to enable the rule and get notices when heredoc syntax is used

It is not limited to only aws resources, if someone is able to contribute recommendations on how to filter on resource block names, while still parsing lexical tokens, I'd welcome it! 

Partially resolves: https://github.com/terraform-linters/tflint-ruleset-aws/issues/36

### DIFF
--- a/docs/rules/aws_policy_heredocs_disallowed.md
+++ b/docs/rules/aws_policy_heredocs_disallowed.md
@@ -1,0 +1,71 @@
+# aws_policy_heredocs_disallowed
+
+This rule ensures that heredocs are not in use across the terraform configuration
+
+## Example
+
+We prefer a jsonencoded configuration of a policy
+```hcl
+  resource "aws_iam_policy" "test" {
+    policy = jsonencode({
+        Statement = [
+            {
+                Action = [
+                    "iam:PassRole"
+                ],
+            }
+        ]
+    })
+  }
+```
+over one with a heredoc:
+```hcl
+resource "aws_iam_policy" "test" {
+  policy = <<EOF
+{
+  "Statement": [
+    {
+      "Action":  [
+        "iam:PassRole"
+      ],
+    }
+  ]
+}
+EOF
+}
+```
+
+
+```
+$ tflint
+1 issue(s) found:
+
+Notice: Avoid Use of HEREDOC syntax for policy documents (aws_policy_heredocs_disallowed)
+
+  on test.tf line 3:
+ 3:   policy = <<EOF
+ 4: {
+
+```
+
+
+## Why
+
+Preference is made to policy documents written with jsonencode() functionality vs the Heredoc functionality.
+
+## How To Fix
+
+convert your heredoc code to json
+```hcl
+  resource "aws_iam_policy" "test" {
+    policy = jsonencode({
+        Statement = [
+            {
+                Action = [
+                    "iam:PassRole"
+                ],
+            }
+        ]
+    })
+  }
+```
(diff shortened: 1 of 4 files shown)


wata727:
> It is not limited to only aws resources, if someone is able to contribute recommendations on how to filter on resource block names, while still parsing lexical tokens, I'd welcome it!

`GetResourceContent` can be used to filter resources. You can pass a range to `hclsyntax.LexConfig`, so you can perform lexical analysis by specifying the range of the target resource expression here.

docs/rules/aws_policy_heredocs_disallowed.md
```diff
@@ -0,0 +1,71 @@
+# aws_policy_heredocs_disallowed
+
+This rule ensures that heredocs are not in use across the terraform configuration
+
+## Example
+
+We prefer a jsonencoded configuration of a policy
```
bendrucker, line 7: "We prefer" is the wrong subject (who is we?). Rules that assert best practices should either link to upstream documentation or explain why a user should prefer an approach. Avoid introducing personal preferences.

bendrucker: [`data.aws_iam_policy_document`](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/iam_policy_document) provides the most guardrails (schema validation) and should generally be the default. There are valid reasons to use JSON strings, e.g. if copy and pasting a static policy. There is not, as far as I can think of, any case where `jsonencode` is the optimal way to produce an IAM policy.

rules/aws_policy_heredocs_disallowed.go
```diff
@@ -0,0 +1,97 @@
...
+	for name, file := range files {
+		if err := r.checkHeredocs(runner, name, file); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (r *AwsPolicyHeredocsDisallowedRule) checkHeredocs(runner tflint.Runner, filename string, file *hcl.File) error {
```
bendrucker, line 72: As already noted, this needs to filter on AWS resources. But beyond that, it is not correct to simply forbid all heredocs in AWS resources. Take this perfectly valid resource for example:

```tf
resource "aws_s3_bucket_object" "readme" {
  bucket = "your_bucket_name"
  key    = "README.md"
  source = "path/to/file"

  content = <<-EOF
    # Hello world!

    I am a markdown document, embedded in a heredoc
  EOF
}
```

Only resource attributes that accept an IAM policy can be lexed for heredoc tokens. This is probably the hardest part of implementing the rule. You might be able to look at name patterns for those attributes but it's also possible you'd have to keep a static `map[string][]string` of resource names to attributes that accept IAM policies.

docs/rules/aws_policy_heredocs_disallowed.md
bendrucker: Arguably this rule should warn on `jsonencode` too and be named positively `aws_iam_policy_document_data_source` instead of focusing on what not to do. Heredocs are not so different than `file()` or `templatefile()`.

bendrucker (changes requested):
Are you interested in finishing this PR? Left some comments explaining some significant unsolved challenges. If it feels like too much I understand and will close this out.

wata727:
Closing this PR due to no recent activity. Feel free to reopen if you are interested in continuing this work. Thanks!

Issue #35, opened 2021-01-04
feat: Generic rule to prevent certain dangerous resources like aws_iam_policy_attachment


nitrocode:
There is a large warning for the resource [`aws_iam_policy_attachment`](https://www.terraform.io/docs/providers/aws/r/iam_policy_attachment.html) to use [`aws_iam_role_policy_attachment`](https://www.terraform.io/docs/providers/aws/r/iam_role_policy_attachment.html) instead.

I'd like to prevent the first resource from getting into our infrastructure with an appropriate tflint rule. We could make it specific to this resource but I'd prefer a more generic one so if we find a future resource, we can simply add it to a list within the rule.

```lang-hcl
rule "terraform_denylisted_resources" {
  enabled = true

  resources = [
    "aws_iam_policy_attachment",
    "google_organization_iam_binding",
  ]

  # in case some upstream module refuses to remove xyz resource 
  # and you don't want to manage a fork
  allow_in_modules = true
}
```

Edit: @mveitas put in PR https://github.com/terraform-linters/tflint/pull/769

m00lecule:
@mveitas are there still chances, that this rule will be implemented?

PR #384, opened 2022-10-10
Add aws_iam_policy_attachment_has_alternatives rule

metafeather:
Add `aws_iam_policy_attachment_has_alternatives` rule.

This rule warns when there are recommended alternatives to the resource, and is _disabled_ by default

This recommendation is documented by HashiCorp in the AWS Provider documentation due to  common misuse: https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/iam_policy_attachment

In this particular case the blast radius of misuse can be severe across an entire organisation account and cause production outages due to the unexpected removal of permissions.

After reviewing past issues (#35) and PR's ([tflint:#769](https://github.com/terraform-linters/tflint/pull/769)) I think that adding individual rules informed by HashiCorp recommendations is the most practical approach for community maintenance and this PR could be used as an approachable template for further rules without deep knowledge of the project (I used #355)

Thankyou for your consideration.

### DIFF
--- a/rules/aws_iam_policy_attachment_has_alternatives.go
+++ b/rules/aws_iam_policy_attachment_has_alternatives.go
@@ -1,0 +1,74 @@
+package rules
+
+import (
+	"github.com/terraform-linters/tflint-plugin-sdk/hclext"
+	"github.com/terraform-linters/tflint-plugin-sdk/tflint"
+	"github.com/terraform-linters/tflint-ruleset-aws/project"
+)
+
+// AwsIAMPolicyAttachmentHasAlternativesRule warns that the resource has alternatives recommended
+type AwsIAMPolicyAttachmentHasAlternativesRule struct {
+	tflint.DefaultRule
+
+	resourceType  string
+	attributeName string
+}
+
+// AwsIAMPolicyAttachmentHasAlternativesRule returns new rule with default attributes
+func NewAwsIAMPolicyAttachmentHasAlternativesRule() *AwsIAMPolicyAttachmentHasAlternativesRule {
+	return &AwsIAMPolicyAttachmentHasAlternativesRule{
+		resourceType:  "aws_iam_policy_attachment",
+		attributeName: "name",
+	}
+}
+
+// Name returns the rule name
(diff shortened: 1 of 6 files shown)


rules/aws_iam_policy_attachment_has_alternatives_test.go
```diff
@@ -0,0 +1,51 @@
...
+					Rule:    NewAwsIAMPolicyAttachmentHasAlternativesRule(),
+					Message: "Consider aws_iam_role_policy_attachment, aws_iam_user_policy_attachment, or aws_iam_group_policy_attachment instead.",
+					Range: hcl.Range{
+						Filename: "resource.tf",
+						Start:    hcl.Pos{Line: 3, Column: 9},
+						End:      hcl.Pos{Line: 3, Column: 26},
+					},
+				},
+			},
+		},
```
wata727, line 37: It would be nice to have a fixed case with no warnings.

rules/aws_iam_policy_attachment_has_alternatives.go
```diff
@@ -0,0 +1,74 @@
...
+func NewAwsIAMPolicyAttachmentHasAlternativesRule() *AwsIAMPolicyAttachmentHasAlternativesRule {
+	return &AwsIAMPolicyAttachmentHasAlternativesRule{
+		resourceType:  "aws_iam_policy_attachment",
+		attributeName: "name",
+	}
+}
+
+// Name returns the rule name
+func (r *AwsIAMPolicyAttachmentHasAlternativesRule) Name() string {
+	return "aws_iam_policy_attachment_has_alternatives"
```
wata727, line 27: How about a name like `aws_iam_policy_attachment_deprecated`?

wata727: Or `aws_iam_policy_attachment_exclusive_attachment` may be better.

I prefer names that are descriptive of what issue we are warning about. Additionally, the prefix should preferably match the resource name. What do you think?

bendrucker:
Are you interested in finishing this PR?

kayman-mk:
@bendrucker Was this implemented in the meantime?

PR #405, opened 2022-12-02
Add assume role configuration in Config struct

kaito3desuyo:
My Terraform project has the following folder structure:
```
<root>
├ environments
│　├ dev
│　├ stg
│　├ prod
│　└ <and more...>
├ modules
│　└ <some modules>
├ <some shellscript>
└ .tflint.hcl 
```

And I'm running tflint with the following shell script.
```sh
#!/bin/bash
PATH_DIR_SCRIPT="$(cd "$(dirname "${BASH_SOURCE:-$0}")" && pwd)";

set -euxo pipefail
find . -type f -name "*.tf" -not -path "*/.terraform/*" -exec dirname {} \; \
    | sort -u \
    | while read line; do
        cd $line;
        echo "processing $PWD ..."
        terraform init -backend=false && tflint --config $PATH_DIR_SCRIPT/.tflint.hcl;
        cd -;
      done
```

The problem is that when inspecting the modules in the `modules` folder, the AssumeRole is not possible because there is no provider clause.
Modules in the environments folder work fine because they have a provider clause specified.

Therefore, it would be convenient if AssumeRole can be set in `.tflint.hcl`.
This allows TFLint to run with a consistent Role regardless of whether it is set in the module.

Please consider adding this function.
Thank you.

### DIFF
--- a/aws/config.go
+++ b/aws/config.go
@@ -16,14 +16,14 @@ type Config struct {
 
 func (c *Config) toCredentials() Credentials {
 	return Credentials{
-		AccessKey: 				c.AccessKey,
-		SecretKey: 				c.SecretKey,
-		Region:    				c.Region,
-		Profile:   				c.Profile,
-		CredsFile: 				c.SharedCredentialsFile,
-		AssumeRoleARN: 			c.AssumeRoleARN,
-		AssumeRoleExternalID: 	c.AssumeRoleExternalID,
-		AssumeRolePolicy: 		c.AssumeRolePolicy,
-		AssumeRoleSessionName: 	c.AssumeRoleSessionName,
+		AccessKey:              c.AccessKey,
+		SecretKey:              c.SecretKey,
+		Region:                 c.Region,
+		Profile:                c.Profile,
+		CredsFile:              c.SharedCredentialsFile,
+		AssumeRoleARN:          c.AssumeRoleARN,
+		AssumeRoleExternalID:   c.AssumeRoleExternalID,
+		AssumeRolePolicy:       c.AssumeRolePolicy,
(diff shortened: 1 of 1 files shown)


aws/config.go
```diff
@@ -8,14 +8,22 @@ type Config struct {
 	Region                string `hclext:"region,optional"`
 	Profile               string `hclext:"profile,optional"`
 	SharedCredentialsFile string `hclext:"shared_credentials_file,optional"`
+	AssumeRoleARN         string `hclext:"assume_role_arn,optional"`
+	AssumeRoleExternalID  string `hclext:"assume_role_external_id,optional"`
+	AssumeRolePolicy      string `hclext:"assume_role_policy,optional"`
+	AssumeRoleSessionName string `hclext:"assume_role_session_name,optional"`
```
wata727, line 14: I think it's easier to understand if you can declare it in the `assume_role` block like Terraform. e.g:

```hcl
plugin "aws" {
  enabled = true

  assume_role {
    role_arn     = "..."
    external_id  = "..."
    policy       = "..."
    session_name = "..."
  }
}
```

Nested structures can be decoded as follows:
https://pkg.go.dev/github.com/terraform-linters/tflint-plugin-sdk@v0.14.0/hclext#example-DecodeBody

wata727 (commented):
This change makes sense to me.

kaito3desuyo:
Thank you for the information.
Added fix to declare in `assume_role` block like Terraform.

aws/config.go
kaito3desuyo: Thank you for the information.
Added fix to declare in assume_role block like Terraform.

wata727 (approved):
👍

PR #440, opened 2023-01-03
[WIP] Add preset filtering

marinalimeira:
This is WIP.

Validating what it takes to add preset validation to the project.


wata727:
Thank you for working on this. What is the status of this?
Are you planning to integrate the https://github.com/gruntwork-io/tflint-ruleset-aws-cis into the AWS ruleset?

marinalimeira:
Hello!

The rules filter was a crucial feature to us, our use case needs executing a subset of the rules - not the all rules -  which isn't supported in this ruleset yet (but the terraform ruleset does).

I started to work on this functionality but due to time constraints, it was quicker to launch a new repo with our custom rules instead of continue working on this PR.

marinalimeira:
I just updated this PR to remove the CIDR block check, given it was released in https://github.com/gruntwork-io/tflint-ruleset-aws-cis.

aws/ruleset.go
```diff
@@ -29,14 +33,43 @@ func (r *RuleSet) ApplyConfig(body *hclext.BodyContent) error {
 		return nil
 	}
 
+	preset := map[string]bool{}
```
marinalimeira, line 36: I couldn't make this work. The tests weren't passing. I can get back to this but it will take me some weeks to be able to focus on this change. I think this feature would be a nice addition to the repo!

wata727:
Thank you for your explanation. However, merging this change may be a bit difficult as there are no good preset candidates at this time.

I think it's appropriate that the CIS ruleset be released as another plugin, as has already been done. In this way, if you want to use a new preset, I think in most cases it would be better to release them as another plugin.

marinalimeira:
Got it! I will chose this PR then. Thank you for the reviews.

PR #455, opened 2023-02-23
`aws_acm_certificate`: fix false positive for private CA

bendrucker:
Correctly handles `certificate_authority_arn` for an `aws_acm_certificate`. Closes https://github.com/terraform-linters/tflint-ruleset-aws/issues/453. 

This ran up against issues with the test generator, which fails to account for `truncateLongMessage` in the tests. This is marked as a draft so that this can be fixed and then this change can be rebased on top.

PR #456, opened 2023-02-23
models: allow tests to specify multiple cases

bendrucker:
In #455 I found the model tests to be awkward to work with. `ok` and `ng` are confusing (no good?). Multiple test cases are not supported for the same attribute. 

Instead:

```hcl
test "aws_acmpca_certificate_authority" "type" {
  valid   = ["SUBORDINATE"]
  invalid = ["ORDINATE"]
}
```

* Clear attribute naming
* Multiple cases

I could break out the valid and invalid tests into separate sub-tests for easier identification but the naming seems like a waste. If anything, we shouldn't be allowing the SDK framework to call `t.Fatal`. We should be diffing the issues and printing the location of the mapping file in the error if there's a diff. Treating the generated test like a traditional Go test that the user would edit directly if failing is not the intended workflow. 

Separately, there are also bugs, where the tests don't account for truncation performed in the generated rules (see #455). 

Some diffs here (missing double quotes) are unintended consequences of refactoring the template and its calling code.

### DIFF
--- a/rules/models/aws_acmpca_certificate_authority_invalid_type_test.go
+++ b/rules/models/aws_acmpca_certificate_authority_invalid_type_test.go
@@ -10,30 +10,27 @@ import (
 
 func Test_AwsAcmpcaCertificateAuthorityInvalidTypeRule(t *testing.T) {
 	cases := []struct {
-		Name     string
 		Content  string
 		Expected helper.Issues
 	}{
 		{
-			Name: "It includes invalid characters",
 			Content: `
 resource "aws_acmpca_certificate_authority" "foo" {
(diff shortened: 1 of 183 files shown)


bendrucker:
Arguably this is still the wrong API. This was easy to programmatically rewrite into using regexes, but a better API would be:

```tf
test "aws_acmpca_certificate_authority" "type" {
  case {
    value = "SUBORDINATE"
  }

  case {
    value = "ORDINATE"
    valid = false
  }
}
```

Failing tests should not be required. This also opens up an option to add an optional `name` or similar to the `case {}` block, to describe the value.

PR #499, opened 2023-06-16
No s3 global endpoint rule

AleksaC:
Hi, I have an idea for a new rule. The idea is to disallow usage of [s3 global endpoints](https://docs.aws.amazon.com/AmazonS3/latest/userguide/VirtualHosting.html#deprecated-global-endpoint) which have been deprecated. The most notable issue caused by this is cloudfront redirecting to the regional endpoint for some time after the bucket has been created, as described [here](https://stackoverflow.com/questions/38735306/aws-cloudfront-redirecting-to-s3-bucket/58423033#58423033).

The [`aws_s3_bucket`](https://registry.terraform.io/providers/hashicorp/aws/5.4.0/docs/resources/s3_bucket) resource returns the global endpoint in [`bucket_domain_name`](https://registry.terraform.io/providers/hashicorp/aws/5.4.0/docs/resources/s3_bucket#bucket_domain_name) output which is what my current implementation of the rule catches. However I'd also look for strings that contain something like `bucketname.s3.amazonaws.com`

This is a rough but working implementation of the rule, so I'm opening a draft PR to get your feedback before doing additional work. If you think this is a useful rule, I'll cover additional cases, clean up the code and add documentation.


rules/aws_s3_no_global_endpoint.go
```diff
@@ -0,0 +1,68 @@
...
+}
+
+// Name returns the rule name
+func (r *AwsS3NoGlobalEndpointRule) Name() string {
+	return "aws_acm_certificate_lifecycle"
+}
+
+// Enabled returns whether the rule is enabled by default
+func (r *AwsS3NoGlobalEndpointRule) Enabled() bool {
+	return true
```
bendrucker, line 34: Given that the attribute isn't deprecated in the provider itself, this definitely should not be enabled by default. Terraform will automatically print warnings when deprecated attributes are referenced so if there's truly no need for this today except for exceptional backwards compatibility cases it should probably be deprecated upstream too. Not a requirement for merging this rule but worth kicking off given that it's a 1 line change.

rules/aws_s3_no_global_endpoint.go
```diff
@@ -0,0 +1,68 @@
...
+		vars := expr.Variables()
+
+		if len(vars) == 0 {
+			return nil
+		}
+
+		// is this ever greater than 0
+		v := vars[0]
+
+		if v.RootName() == "aws_s3_bucket" && len(v) == 3 && v[2].(hcl.TraverseAttr).Name == "bucket_domain_name" {
```
bendrucker, line 60: This should be more robust, `ReferencesInExpr` may be helpful:

https://github.com/terraform-linters/tflint-plugin-sdk/blob/b875e9279047ce4902eec3d03c165f32d8edfe26/terraform/lang/references.go#L41-L60

bendrucker (changes requested):
Seems like a reasonable rule. Happy to give this another pass when it's closer to ready. For now I've made some high level notes. I'd encourage you to think of test cases that will break your parser. Things like using count/for_each.

bendrucker:
Are you interested in finishing this?

AleksaC:
Yes I am. This turned out to be harder to implement than I initially thought and things kept coming up for me so I didn't have time to take a proper look at it. I should have more time in the following few days so will try to pick it up again.

PR #554, opened 2023-09-25
`s3_bucket_name`: add length validation

davimmt:
From [this discussion](https://github.com/terraform-linters/tflint-ruleset-aws/discussions/539#discussion-5595579).

image


rules/aws_s3_bucket_name_length.go
```diff
@@ -0,0 +1,84 @@
...
+		Attributes: []hclext.AttributeSchema{
+			{Name: r.attributeName},
+		},
+	}, nil)
+	if err != nil {
+		return err
+	}
+
+  bucketNameMinLength := 3
+  bucketNameMaxLength := 63
```
bendrucker, line 60: Can you avoid hardcoding this and find somewhere to reference it? AWS has a lot of schemas for their API that should specify this somewhere...

bendrucker:
Seems like this could be added to the existing `aws_s3_bucket_name` rule. Ideally that would have been `aws_s3_bucket_name_prefix` and `aws_s3_bucket_name` would just enforce the [official rules](https://docs.aws.amazon.com/AmazonS3/latest/userguide/bucketnamingrules.html). In addition to the length limit, the rule should enforce all the other requirements.

rules/aws_s3_bucket_name_length.go
bendrucker: Hmm, seems not:

https://github.com/aws/aws-sdk-go/blob/2ac2bb7f4efd6406c91b30e668024928ba8a63bc/models/apis/s3/2006-03-01/api-2.json#L1350

rules/aws_s3_bucket_name_length.go
```diff
@@ -0,0 +1,84 @@
...
+		attribute, exists := resource.Body.Attributes[r.attributeName]
+		if !exists {
+			continue
+		}
+
+		err := runner.EvaluateExpr(attribute.Expr, func (name string) error {
+      if len(name) < bucketNameMinLength || len(name) > bucketNameMaxLength {
+				runner.EmitIssue(
+					r,
+          fmt.Sprintf(`Bucket name "%s" length must be within %d - %d character range`, name, bucketNameMinLength, bucketNameMaxLength),
```
bendrucker, line 72:
```suggestion
          fmt.Sprintf("Bucket name %q must be between %d and %d characters", name, bucketNameMinLength, bucketNameMaxLength),
```

rules/aws_s3_bucket_name_length.go
davimmt: Yeah, it seems rather bad to hardcode, but I also haven't found somewhere to reference it from.

davimmt:
> Seems like this could be added to the existing `aws_s3_bucket_name` rule. Ideally that would have been `aws_s3_bucket_name_prefix` and `aws_s3_bucket_name` would just enforce the [official rules](https://docs.aws.amazon.com/AmazonS3/latest/userguide/bucketnamingrules.html). In addition to the length limit, the rule should enforce all the other requirements.

I thought about it and I think it's probably the right thing to do; however, for what I've seen, the `aws_s3_bucket_name` rule requires an explicit regex configuration and it's not enabled by default (as the bucket name length enforcement should always be), so I decided to make it separate.

bendrucker:
Right, so:

> requires an explicit regex configuration

That would need to become optional

> it's not enabled by default

Then this can happen

davimmt:
Yeah, that's right. I was hoping your feedback on this matter, if you agreed on using the same rule. I will work on it, then.

davimmt:
Just checking indentation now, it's weird. I think maybe some misbehaviour on my text editor conf as I don't program in Go.

davimmt:
Seems all right.

image

Have you found anything about the min and max numbers from the API?

Also, your opinion on [this](https://github.com/terraform-linters/tflint-ruleset-aws/pull/554/commits/c20500758e834f8781a87bb8c4469befe0f360ac#diff-69ddc73ffdcaaf0c7b2652ca7e9f8a6da557c808fea56912c0b4fea7b3f0844eL44-L47).

bendrucker:
Thanks, will review and respond within the next 1-2 weeks (traveling)

davimmt:
We could work with this, maybe:

https://github.com/aws/aws-sdk-go/blob/0f30069a590dd0a55ef0375fd9cb0d3b40284cd9/models/apis/cloudformation/2010-05-15/api-2.json#L3338-L3343

bendrucker (approved):
Thanks for your patience!

PR #569, opened 2023-11-02
`s3_bucket_name`: fix docs

davimmt:
Checking the enabled default box:

https://github.com/terraform-linters/tflint-ruleset-aws/blob/master/rules/aws_s3_bucket_name.go#L39C1-L42C2

Removing the `specified` because that's not required by the rule anymore.


bendrucker (changes requested):
Update docs/rules/README.md.tmpl too to resolve the diff error

davimmt (commented):
Done

PR #571, opened 2023-11-06
`s3_bucket_name`: add all documented naming rules

davimmt:
From [this discussion](https://github.com/terraform-linters/tflint-ruleset-aws/discussions/570).

Also an alternative to remove hardcoded values from [this past issue](https://github.com/terraform-linters/tflint-ruleset-aws/pull/554#discussion_r1336202012).

### DIFF
--- a/rules/aws_s3_bucket_name.go
+++ b/rules/aws_s3_bucket_name.go
@@ -81,7 +81,7 @@ func (r *AwsS3BucketNameRule) Check(runner tflint.Runner) error {
 	}{
 		{
 			Regexp: *regexp.MustCompile(fmt.Sprintf(
-				"^(.{0,%d})?(.{%d,})?$",
+				"^.{0,%d}$|.{%d,}$",
 				bucketNameMinLength-1,
 				bucketNameMaxLength+1,
 			)),
@@ -96,12 +96,8 @@ func (r *AwsS3BucketNameRule) Check(runner tflint.Runner) error {
 			Description: "Bucket names can consist only of lowercase letters, numbers, dots (.), and hyphens (-).",
 		},
 		{
-			Regexp:      *regexp.MustCompile("^[^a-z0-9]"),
-			Description: "Bucket names must begin with a lowercase letter or number.",
-		},
-		{
-			Regexp:      *regexp.MustCompile("[^a-z0-9]$"),
-			Description: "Bucket names must end with a lowercase letter or number.",
+			Regexp:      *regexp.MustCompile("^[^a-z0-9]|[^a-z0-9]$"),
+			Description: "Bucket names must begin and end with a lowercase letter or number.",
 		},
 		{
 			Regexp:      *regexp.MustCompile("\\.\\."),
@@ -112,7 +108,7 @@ func (r *AwsS3BucketNameRule) Check(runner tflint.Runner) error {
 			Description: "Bucket names must not start with the prefix 'xn--'.",
 		},
 		{
-			Regexp:      *regexp.MustCompile("^(sthree-dwa|sthree-configurator21)"),
+			Regexp:      *regexp.MustCompile("^(sthree-|sthree-configurator)"),
 			Description: "Bucket names must not start with the prefix 'sthree-' and the prefix 'sthree-configurator'.",
 		},
 		{
--- a/rules/aws_s3_bucket_name.go
+++ b/rules/aws_s3_bucket_name.go
@@ -2,6 +2,7 @@ package rules
 
 import (
 	"fmt"
+	"net"
 	"regexp"
 	"strings"
 
@@ -73,6 +74,57 @@ func (r *AwsS3BucketNameRule) Check(runner tflint.Runner) error {
 	bucketNameMinLength := 3
 	bucketNameMaxLength := 63
 
+	// https://docs.aws.amazon.com/AmazonS3/latest/userguide/bucketnamingrules.html
+	regexRules := []struct {
+		Regexp      regexp.Regexp
+		Description string
+	}{
+		{
+			Regexp: *regexp.MustCompile(fmt.Sprintf(
+				"^(.{0,%d})?(.{%d,})?$",
+				bucketNameMinLength-1,
+				bucketNameMaxLength+1,
+			)),
+			Description: fmt.Sprintf(
+				"Bucket names must be between %d (min) and %d (max) characters long.",
+				bucketNameMinLength,
+				bucketNameMaxLength,
+			),
+		},
+		{
+			Regexp:      *regexp.MustCompile("[^a-z0-9\\.\\-]"),
+			Description: "Bucket names can consist only of lowercase letters, numbers, dots (.), and hyphens (-).",
+		},
+		{
+			Regexp:      *regexp.MustCompile("^[^a-z0-9]"),
+			Description: "Bucket names must begin with a lowercase letter or number.",
+		},
+		{
+			Regexp:      *regexp.MustCompile("[^a-z0-9]$"),
+			Description: "Bucket names must end with a lowercase letter or number.",
+		},
+		{
+			Regexp:      *regexp.MustCompile("\\.\\."),
+			Description: "Bucket names must not contain two adjacent periods.",
+		},
+		{
+			Regexp:      *regexp.MustCompile("^xn--"),
+			Description: "Bucket names must not start with the prefix 'xn--'.",
+		},
+		{
+			Regexp:      *regexp.MustCompile("^(sthree-dwa|sthree-configurator21)"),
+			Description: "Bucket names must not start with the prefix 'sthree-' and the prefix 'sthree-configurator'.",
+		},
+		{
+			Regexp:      *regexp.MustCompile("-s3alias$"),
+			Description: "Bucket names must not end with the suffix '-s3alias'.",
+		},
+		{
+			Regexp:      *regexp.MustCompile("--ol-s3$"),
+			Description: "Bucket names must not end with the suffix '--ol-s3'.",
+		},
+	}
+
 	for _, resource := range resources.Blocks {
 		attribute, exists := resource.Body.Attributes[r.attributeName]
 		if !exists {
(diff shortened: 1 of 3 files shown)


rules/aws_s3_bucket_name.go
```diff
@@ -51,6 +64,68 @@ func (r *AwsS3BucketNameRule) Link() string {
 	return project.ReferenceLink(r.Name())
 }
 
+// Return regex bucket naming rules
```
bendrucker, line 67:
```suggestion
// rules returns a list of objects with regular expressions enforcing a name rule and a description of the rule
```

rules/aws_s3_bucket_name.go
```diff
@@ -51,6 +64,68 @@ func (r *AwsS3BucketNameRule) Link() string {
...
+	defer r.Body.Close()
+
+	bucket := Bucket{}
+	err = json.NewDecoder(r.Body).Decode(&bucket)
+	if err != nil {
+		return nil, err
+	}
+
+	// https://docs.aws.amazon.com/AmazonS3/latest/userguide/bucketnamingrules.html
+	return []map[string]string{
```
bendrucker, line 85: Use `struct` when fields are known and reserve `map` for arbitrary keys. Avoid unnecessary shorthand and rely on the fact that defined keys allows quick auto-completion.

```go
type struct S3BucketNameRule {
  Regexp regexp.Regexp
  Description string
}
```

If you take out the network component you don't need to return an error here so it's unclear that `rules()` is even needed as its own function. So you could potentially just define this as inline anonymous struct type:

```go
rules := []struct{
  Regexp regexp.Regexp
  Description string
}{
  {
  // ...
  }
}
```

bendrucker: Additionally, if you take out the network component it should be simpler to define these as a global variable and use `regexp.MustCompile`. It's better to fail at the initialization level and effectively break the build than risk an unexpected runtime error.

rules/aws_s3_bucket_name.go
```diff
@@ -100,12 +177,18 @@ func (r *AwsS3BucketNameRule) Check(runner tflint.Runner) error {
...
-				)
+			for _, rule := range rules {
+				regex, err := regexp.MatchString(rule["regx"], name)
+				if err != nil {
+					return fmt.Errorf("invalid regex (%s): %w", rule["regx"], err)
+				}
+				if regex {
+					runner.EmitIssue(
+						r,
+						fmt.Sprintf(`Bucket name %q does not match regex %q: %q`, name, rule["regx"], rule["desc"]),
```
bendrucker, line 188: Reverse the order to description then regex. The description is for humans, whereas the regex is a machine-interpreted language, offered as an additional diagnostic aid. In some cases the regexes are quite long and push the description far to the right. A rough idea:

```go
fmt.Sprintf("%s (name: %q, regex: %q)", rule.Description, name, rule.Regexp.String())
```

rules/aws_s3_bucket_name.go
```diff
@@ -51,6 +64,68 @@ func (r *AwsS3BucketNameRule) Link() string {
...
+		{
+			"regx": "\\.\\.",
+			"desc":  "Bucket names must not contain two adjacent periods.",
+		},
+		{
+			"regx": "^\\b((25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)(\\.|$)){4}\\b$",
+			"desc":  "Bucket names must not be formatted as an IP address (for example, 192.168.5.4).",
+		},
+		{
+			"regx": "^(xn--|sthree-dwa|sthree-configurator21)",
```
bendrucker, line 119: Avoid merging these together as to keep the descriptions from the official AWS docs. Same for suffixes.

rules/aws_s3_bucket_name.go
```diff
@@ -51,6 +64,68 @@ func (r *AwsS3BucketNameRule) Link() string {
...
+		{
+			"regx": "[^a-z0-9]$",
+			"desc":  "Bucket names must end with a lowercase letter or number.",
+		},
+		{
+			"regx": "\\.\\.",
+			"desc":  "Bucket names must not contain two adjacent periods.",
+		},
+		{
+			"regx": "^\\b((25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)(\\.|$)){4}\\b$",
```
bendrucker, line 115: Any citation for this? A simple `^(\d{1,3}\.){3}\d{1,3}$` needs none but there is much more logic here.

bendrucker (changes requested):

rules/aws_s3_bucket_name.go
davimmt: 192.168.0.1 -> Not OK
393.388.2.321 -> OK

rules/aws_s3_bucket_name.go
davimmt: OK, so:
```go		
{
  Regexp:      *regexp.MustCompile("^xn--"),
  Description: "Bucket names must not start with the prefix 'xn--'.",
},
{
  Regexp:      *regexp.MustCompile("^(sthree-dwa|sthree-configurator21)"),
  Description: "Bucket names must not start with the prefix 'sthree-' and the prefix 'sthree-configurator'.",
},
{
  Regexp:      *regexp.MustCompile("-s3alias$"),
  Description: "Bucket names must not end with the suffix '-s3alias'.",
},
{
  Regexp:      *regexp.MustCompile("--ol-s3$"),
  Description: "Bucket names must not end with the suffix '--ol-s3'.",
},

```

rules/aws_s3_bucket_name.go
davimmt: I'd remove the function and just add this  inline inside _Check_:
```go
bucketNameMinLength := 3
bucketNameMaxLength := 63

// https://docs.aws.amazon.com/AmazonS3/latest/userguide/bucketnamingrules.html
rules := []struct {
  Regexp      regexp.Regexp
  Description string
}{
  {
    Regexp: *regexp.MustCompile(fmt.Sprintf(
      "^(.{0,%d})?(.{%d,})?$",
      bucketNameMinLength-1,
      bucketNameMaxLength+1,
    )),
    Description: fmt.Sprintf(
      "Bucket names must be between %d (min) and %d (max) characters long.",
      bucketNameMinLength,
      bucketNameMaxLength,
    ),
  },
  {
    Regexp:      *regexp.MustCompile("[^a-z0-9\\.\\-]"),
    Description: "Bucket names can consist only of lowercase letters, numbers, dots (.), and hyphens (-).",
  },

  ...

}
```
Or is it preferable to write this another place?

And then:
```go
regex, _ := regexp.MatchString(rule.Regexp.String(), name)
if regex {
  runner.EmitIssue(
    r,
    fmt.Sprintf(`%s (name: %q, regex: %q)`, rule.Description, name, rule.Regexp.String()),
    attribute.Expr.Range(),
  )
}
```

rules/aws_s3_bucket_name.go
```diff
@@ -51,6 +64,68 @@ func (r *AwsS3BucketNameRule) Link() string {
 	return project.ReferenceLink(r.Name())
 }
 
+// Return regex bucket naming rules
+func rules() ([]map[string]string, error) {
+	// Get bucket naming rule based on length from AWS SDK API models
+	r, err := http.Get(
+		"https://raw.githubusercontent.com/aws/aws-sdk-go/main/models/apis/cloudformation/2010-05-15/api-2.json",
```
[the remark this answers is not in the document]
davimmt: That would require project and language knowledge wich I don't currently have. But I'd be happy to do it sometime later, when I have the proper skills. I also talked to some AWS team members and they don't seem like exposing these validations in their api models amytime soon.

rules/aws_s3_bucket_name.go
bendrucker: Yep!

rules/aws_s3_bucket_name.go
bendrucker: Inline is fine. Avoid `regexp.MatchString` in favor of compiling the regex first and then testing the string separately.

rules/aws_s3_bucket_name.go
bendrucker: Sure I get it, but it's still a complicated regex. An alternative:

https://pkg.go.dev/net#ParseIP

If it returns non nil and calling `To4` on the result is also non-nil, you know it's an IPv4 address. That code is easier to read than a regular expression.

davimmt: It'd have to be a separate check other than regex, then.

davimmt:
```go
nameAsIP := net.ParseIP(name)
if nameAsIP != nil && net.IP.To4(nameAsIP) != nil {
  runner.EmitIssue(
    r,
    fmt.Sprintf(`Bucket names must not be formatted as an IP address. (name: %q)`, name),
    attribute.Expr.Range(),
  )
}
```

rules/aws_s3_bucket_name.go
bendrucker: This isn't needed:

```go
regex, err = regexp.Compile(rule.Regexp.String())
```

`rule.Regexp` is itself already a compiled `regexp.Regexp` object. I only mean to avoid the `regexp.MatchString` function which accepts a string regex and string value and both compiles the regex and matches the string. Calling the `MatchString` _method_ on an already compiled regex is fine.

I'm happy to tweak this a bit to fine tune where in the program lifecycle the regexes are compiled.

rules/aws_s3_bucket_name.go
bendrucker: Yep, that looks good to me!

rules/aws_s3_bucket_name.go
davimmt: So the for loop already sets the right compiled regex from the current iterator item?
```go
for _, rule := range regexRules {
  if rule.Regexp.MatchString(name) {
    runner.EmitIssue(
      r,
      fmt.Sprintf(`%s (name: %q, regex: %q)`, rule.Description, name, rule.Regexp.String()),
      attribute.Expr.Range(),
    )
  }
}
```

bendrucker: Correct!

davimmt:
If everything's OK, I'll only need to write a test for each item in `aws_s3_bucket_name_test.go`. These changes requires additional testing in `integration/` too?

bendrucker:
Unit testing this rule should be fine, no integration tests needed.

rules/aws_s3_bucket_name.go
```diff
@@ -81,7 +81,7 @@ func (r *AwsS3BucketNameRule) Check(runner tflint.Runner) error {
 	}{
 		{
 			Regexp: *regexp.MustCompile(fmt.Sprintf(
-				"^(.{0,%d})?(.{%d,})?$",
+				"^.{0,%d}$|.{%d,}$",
```
davimmt, line 84: Found out about the alternation operator in Go and thought it was easier to read and more go-friendly.

rules/aws_s3_bucket_name.go
```diff
@@ -96,12 +96,8 @@ func (r *AwsS3BucketNameRule) Check(runner tflint.Runner) error {
...
 		},
 		{
-			Regexp:      *regexp.MustCompile("^[^a-z0-9]"),
-			Description: "Bucket names must begin with a lowercase letter or number.",
-		},
-		{
-			Regexp:      *regexp.MustCompile("[^a-z0-9]$"),
-			Description: "Bucket names must end with a lowercase letter or number.",
+			Regexp:      *regexp.MustCompile("^[^a-z0-9]|[^a-z0-9]$"),
+			Description: "Bucket names must begin and end with a lowercase letter or number.",
```
davimmt, line 100: The alternation operator could also be used here to merge these two rules in one, just as the docs.

bendrucker:
Thanks! Will do another review and edit as needed later this week.

bendrucker (approved):

PR #574, opened 2023-11-16
Update aws_lambda_function_invalid_runtime.go to reflect current runt…

glitchcowboy:
…imes supported as of 16 NOV 2023

https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html


glitchcowboy:
Apologies if I'm approaching this wrong, I'm not familiar with go, but I am tired of tflint incorrectly asserting that python3.10 is an invalid runtime:). I'm not sure what the generated-code/checks is telling me.

bendrucker:
> I'm not sure what the generated-code/checks is telling me.

https://github.com/terraform-linters/tflint-ruleset-aws/pull/574/files#diff-f0826d6752496fc6e3733c277be445a4d54c2c3937c1b32613d5fb36eb602e8dR1

Additionally, you'll note that Python 3.10 is already in the list, since it's generated from AWS API docs:

https://github.com/terraform-linters/tflint-ruleset-aws/pull/574/files#diff-f0826d6752496fc6e3733c277be445a4d54c2c3937c1b32613d5fb36eb602e8dL56

So for the runtime you're interested moving it to a different place in the slice is a noop. And then you've incorrectly removed a handful of valid runtimes which is the reason we prefer generated rules from API specs over ones interpreted by humans from documentation.

For general Q&A, [Discussions](https://github.com/terraform-linters/tflint-ruleset-aws/discussions) are the place. But I'm guessing you're just running an outdated version and could fix your problem accordingly.

glitchcowboy:
Right on line 1.  Wow.  Well thanks for explaining.  I'm curious though why the AWS API spec would provide a list that differs so greatly from their published docs.  There are a bunch of versions in the generated slice that are not on AWS doc of supported runtimes, and others that aren't there that should be.

bendrucker:
You've omitted the deprecated runtimes for one. Those are still valid, but new Lambdas cannot be created using them. There's a separate hand-maintained rule (since there's no source of structured data for deprecated runtimes) that will reject those:

https://github.com/terraform-linters/tflint-ruleset-aws/blob/e3d27b60c3f38ba859a5d9930d34680bc83d2bdd/rules/aws_lambda_function_deprecated_runtime.go#L29-L60

Node.js 20 is brand new and may not have landed in the SDK yet, or if so our automated update PR hasn't happened yet. 

AFAICT the rule and the AWS docs match exactly.

PR #633, opened 2024-05-01
New rule: aws_provider_missing_tags

bootswithdefer:
Adds an `aws_provider_missing_tags` rule, which requires specific tags for AWS provider default_tags blocks.  The `aws_resource_missing_tags` rule is not sufficient, it enforces the set of tags that end up on the resource, but it doesn't enforce the practice of applying them via default_tags instead of resource tags.  Using default tags results in more DRY terraform.

I would see this being used in conjunction with `aws_resource_missing_tags`, where `aws_provider_missing_tags` is used for common tags and `aws_resource_missing_tags` is used for more resource-specific tags (like Name).


rules/aws_provider_missing_tags.go
```diff
@@ -0,0 +1,160 @@
...
+				return err
+			}
+			continue
+		}
+
+		for _, block := range provider.Body.Blocks {
+			var providerTags []string
+			attr, ok := block.Body.Attributes[providerTagsAttributeName]
+			if !ok {
+				r.emitIssue(runner, providerTags, config, provider.DefRange)
```
wata727, line 118: `block.DefRange` would be better than `provider.DefRange`.

rules/aws_provider_missing_tags.go
```diff
@@ -0,0 +1,160 @@
...
+				providerTags = keys
+				return nil
+			}, nil)
+
+			if err != nil {
+				return nil
+			}
+
+			// Check tags
+			r.emitIssue(runner, providerTags, config, provider.DefRange)
```
wata727, line 140: `attr.Range` would be better than `provider.DefRange`.

docs/rules/aws_provider_missing_tags.md
```diff
@@ -0,0 +1,47 @@
...
+
+Notice: The provider is missing the following tags: "Bar", "Foo". (aws_provider_missing_tags)
+
+  on test.tf line 1:
+   1: provider "aws" {
+```
+
+## Why
+
+You want to set a standardized set of tags for your AWS resources via the provider default tags.
```
wata727, line 43: I think it would be more helpful to mention its use with the `aws_resource_missing_tags` rule in specific scenarios.
At least, this is what I thought about this rule first:

- Why not `aws_resource_missing_tags` rule?
- What is the difference from `aws_resource_missing_tags` rule?
- When should I use this rule?

It's a good idea to include answers to these questions in the "Why" section. Perhaps the explanation in the PR will be helpful.

rules/aws_provider_missing_tags.go
bootswithdefer: There is no block, this case is for when default_tags doesn't exist.

rules/aws_provider_missing_tags.go
```diff
@@ -0,0 +1,160 @@
...
+	"sort"
+	"strings"
+
+	hcl "github.com/hashicorp/hcl/v2"
+	"github.com/terraform-linters/tflint-plugin-sdk/hclext"
+	"github.com/terraform-linters/tflint-plugin-sdk/logger"
+	"github.com/terraform-linters/tflint-plugin-sdk/tflint"
+	"github.com/terraform-linters/tflint-ruleset-aws/project"
+	"github.com/zclconf/go-cty/cty"
+	"golang.org/x/exp/slices"
```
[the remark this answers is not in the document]
bootswithdefer: fixed

rules/aws_provider_missing_tags_test.go
```diff
@@ -0,0 +1,130 @@
...
+}`,
+			Config: `
+rule "aws_provider_missing_tags" {
+  enabled = true
+  tags = ["Bazz", "Fooz"]
+}`,
+			Expected: helper.Issues{
+				{
+					Rule:    NewAwsProviderMissingTagsRule(),
+					Message: "The provider `default` is missing the `default_tags` block",
```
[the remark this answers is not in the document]
bootswithdefer: changed the output, though it's kind of awkward

rules/aws_provider_missing_tags.go
```diff
@@ -0,0 +1,160 @@
...
+	for _, tag := range config.Tags {
+		if !slices.Contains(tags, tag) {
+			missing = append(missing, fmt.Sprintf("%q", tag))
+		}
+	}
+	if len(missing) > 0 {
+		sort.Strings(missing)
+		wanted := strings.Join(missing, ", ")
+		issue := fmt.Sprintf("The provider is missing the following tags: %s.", wanted)
+		runner.EmitIssue(r, issue, location)
```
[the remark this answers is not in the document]
bootswithdefer: done

rules/aws_provider_missing_tags.go
bootswithdefer: changed

docs/rules/aws_provider_missing_tags.md
bootswithdefer: fleshed out the doc

wata727 (approved):
Looks good. Thanks!

max-rocket-internet:
I love this rule, thanks @bootswithdefer 💖

max-rocket-internet:
@wata727 could create a new tag so we can use it? AFAIK we can't add a plugin with a commit ID as the version?

wata727:
v0.32.0 has been released.
https://github.com/terraform-linters/tflint-ruleset-aws/releases/tag/v0.32.0

PR #664, opened 2024-08-02
Add aws_iam_access_key_forbidden rule

HadrienPatte:
Add a new `aws_iam_access_key_forbidden` rule that prevents the creation of IAM access keys through terraform. The rule is disabled by default.

### DIFF
--- a/docs/rules/aws_iam_access_key_forbidden.md
+++ b/docs/rules/aws_iam_access_key_forbidden.md
@@ -1,0 +1,29 @@
+# aws_iam_access_key_forbidden
+
+Disallow using `aws_iam_access_key` resources.
+
+## Example
+
+```hcl
+resource "aws_iam_access_key" "my_key" {
+  user = "my-user"
+}
+```
+
+```
+$ tflint
+1 issue(s) found:
+
+Error: "aws_iam_access_key" resources are forbidden. (aws_iam_access_key_forbidden)
+
+  on main.tf line 1:
+   1: resource "aws_iam_access_key" "my_key" {
+```
+
+## Why
+
+Managing IAM access keys with terraform leads to sensitive secrets being stored in plain text in terraform state files.
+
+## How To Fix
+
+Do not manage IAM access keys with terraform, or use the [pgp_key](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/iam_access_key#pgp_key) argument and protect your backend state file judiciously.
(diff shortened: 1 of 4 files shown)


rules/aws_iam_access_key_forbidden.go
```diff
@@ -0,0 +1,59 @@
...
+	if err != nil {
+		return err
+	}
+
+	for _, resource := range resources.Blocks {
+		runner.EmitIssue(
+			r,
+			"\"aws_iam_access_key\" resources are forbidden.",
+			resource.DefRange,
+		)
```
wata727, line 55: According to the documentation it seems safe if `pgp_key` is set. Is it better to not emit an issue if the `pgp_key` is present?

docs/rules/aws_iam_access_key_forbidden.md
```diff
@@ -0,0 +1,29 @@
...
+
+Error: "aws_iam_access_key" resources are forbidden. (aws_iam_access_key_forbidden)
+
+  on main.tf line 1:
+   1: resource "aws_iam_access_key" "my_key" {
+```
+
+## Why
+
+Managing IAM access keys with terraform leads to sensitive secrets being stored in plain text in terraform state files.
```
wata727, line 25: It seems that the secret is not included in the state when imported.

>  This attribute is not available for imported resources.

https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/iam_access_key#secret

For example, it may be difficult to enforce this rule in use cases where you manually create IAM keys, import them, and manage them with Terraform. What do you think?

wata727 (commented):
Thank you for your contribution!

But I'm wondering if we should have this rule. While I understand the desire to not accidentally include IAM access keys in Terraform state, it's probably not a best practice you can always enable and should be run with your organization's policies. For a requirement like this, the [OPA ruleset](https://github.com/terraform-linters/tflint-ruleset-opa) may be preferable.

That said, this policy may be applicable to many organizations, so I'm not entirely opposed to adding it to the AWS ruleset. I'm just concerned that we'll end up with a lot of rules if we continue to accept various policies as disabled rules by default.

What do you think about this?

HadrienPatte:
Thanks for the review, you're right that preventing the creation of access keys is more a matter of organizational policy than a generic terraform best practice and hence this tflint ruleset is probably not the best place for this kind of rule.

PR #759, opened 2024-11-22
glue: disable connection type and properties checks

### LINKED ISSUE #757: aws_glue_connection_invalid_connection_type missing support BIGQUERY and other values.
## tl;dr;

There are `Error: "BIGQUERY" is an invalid value as connection_type (aws_glue_connection_invalid_connection_type)` even when using terraform-aws-provider v5.56.0 or higher.

## Description of issue

terraform-provider-aws v5.56.0 introduces support for `aws_glue_connection`, adding `AZURECOSMOS`, `AZURESQL`, `BIGQUERY`, `OPENSEARCH`, and `SNOWFLAKE` as valid values for the connection_type argument.

> see: https://github.com/hashicorp/terraform-provider-aws/releases/tag/v5.60.0

However tflint still detects the following error.

```
1 issue(s) found:

Error: "BIGQUERY" is an invalid value as connection_type (aws_glue_connection_invalid_connection_type)

  on main.tf line 19:
  19:   connection_type = "BIGQUERY"
```

It seems that [AwsGlueConnectionInvalidConnectionTypeRule.NewAwsGlueConnectionInvalidConnectionTypeRule](https://github.com/terraform-linters/tflint-ruleset-aws/blob/4ee7ea998cd205afcf23e41f935eacfc2d5596ce/rules/models/aws_glue_connection_invalid_connection_type.go#L23) is missing the definition for `BIGQUERY` and possibly other values. Could it be that we need to add these values to the `connection_type` enum? Is this understanding correct?

## Reproduce code

```terraform
resource "aws_glue_connection" "bigquery_connection" {
  name            = "bigquery-connection"
  connection_type = "BIGQUERY"
  connection_properties = {
    SparkProperties = jsonencode({
      secretId = "<reducted>"
    })
  }
}
```

## Versions

* Terraform: v1.9.7
* terraform-provider-aws: 5.77.0
* tflint: 0.54.0
* tflint-ruleset-aws: 0.35.0

bendrucker:
Relaxes type requirements where the provider resource has diverged from the API models:

https://github.com/hashicorp/terraform-provider-aws/pull/37731

> These types were confirmed by AWS support to be supported but not fully documented. They are also not available in types and enums in various SDKs, including the AWS Go SDK v2 which the Terraform AWS Provider uses.

So in effect, we can't validate these without maintaining a matching list of connection types.

Closes #757


bendrucker:
I'm inclined to just remove the rule, but could also replace the implementation with a noop now that it's not actually generated.

wata727 (approved):
Thanks!

> These types were confirmed by AWS support to be supported but not fully documented

😞

wata727:
>  could also replace the implementation with a noop

Yeah, this is a good approach since it avoids the "rule does not exist" error if compatibility is important. However, for now I agree to simply remove the rule.
In the future, we will need to decide how often we will introduce incompatibilities after the v1 release, and consider a procedure to replace them with noop in the meantime.

PR #786, opened 2024-12-12
feat: add aws_iam_policy_attachment_exclusive_attachment rule

### LINKED ISSUE #35: feat: Generic rule to prevent certain dangerous resources like aws_iam_policy_attachment

kayman-mk:
Add `aws_iam_policy_attachment_exclusive_attachment_rule`.

This rule warns when there are recommended alternatives to the resource, and is disabled by default

This recommendation is documented by HashiCorp in the AWS Provider documentation due to common misuse: https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/iam_policy_attachment

In this particular case the blast radius of misuse can be severe across an entire organization account and cause production outages due to the unexpected removal of permissions.

After reviewing past issues (https://github.com/terraform-linters/tflint-ruleset-aws/issues/35) and PR's (https://github.com/terraform-linters/tflint/pull/769) I think that adding individual rules informed by HashiCorp recommendations is the most practical approach for community maintenance and this PR could be used as an approachable template for further rules without deep knowledge of the project (I used https://github.com/terraform-linters/tflint-ruleset-aws/pull/355)

Thank you for your consideration.

Closes #35


.gitignore
```diff
@@ -1 +1,3 @@
+.idea/
```
bendrucker, line 1: Revert this

docs/rules/aws_iam_policy_attachment_exclusive_attachment.md
```diff
@@ -0,0 +1,35 @@
+# aws_iam_policy_attachment_exclusive_attachment
+
+Consider alternative resources to `aws_iam_policy_attachment`.
```
bendrucker, line 3: Get a little bit of the why up here. The section below can expand but this intro should give the reader a 1-2 sentence summary of what _and_ why.

docs/rules/aws_iam_policy_attachment_exclusive_attachment.md
```diff
@@ -0,0 +1,35 @@
...
+$ tflint
+1 issue(s) found:
+Warning: Consider aws_iam_role_policy_attachment, aws_iam_user_policy_attachment, or aws_iam_group_policy_attachment instead. (aws_iam_policy_attachment_has_alternatives)
+  on template.tf line 2:
+   2:   name "test_attachment"
+```
+
+## Why
+
+The `aws_iam_policy_attachment` resource creates exclusive attachments of IAM policies. Across the entire AWS account, all the users/roles/groups to which a single policy is attached must be declared by a single `aws_iam_policy_attachment` resource. This means that even any users/roles/groups that have the attached policy via any other mechanism (including other Terraform resources) will have that attached policy revoked by this resource. https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/iam_policy_attachment
```
bendrucker, line 31: Hyperlink the first use of the identifier (in backticks) instead of appending the URL at the end

rules/aws_iam_policy_attachment_exclusive_attachment.go
```diff
@@ -0,0 +1,72 @@
...
+
+	for _, resource := range resources.Blocks {
+		attribute, exists := resource.Body.Attributes[r.attributeName]
+		if !exists {
+			continue
+		}
+
+		runner.EmitIssue(
+			r,
+			"Consider aws_iam_role_policy_attachment, aws_iam_user_policy_attachment, or aws_iam_group_policy_attachment instead.",
```
bendrucker, line 62: Include a brief bit of why here (exclusive for whole account)

bendrucker (approved):
Overall looks good! I have a few copy edits and am reviewing via mobile so I'm adding them as comments. You're welcome to address them but I can also just take care of them and then merge in the next few days.

rules/aws_iam_policy_attachment_exclusive_attachment.go
```diff
@@ -0,0 +1,72 @@
...
+func (r *AwsIAMPolicyAttachmentExclusiveAttachmentRule) Check(runner tflint.Runner) error {
+	resources, err := runner.GetResourceContent(r.resourceType, &hclext.BodySchema{
+		Attributes: []hclext.AttributeSchema{{Name: r.attributeName}},
+	}, nil)
+	if err != nil {
+		return err
+	}
+
+	for _, resource := range resources.Blocks {
+		attribute, exists := resource.Body.Attributes[r.attributeName]
```
bendrucker, line 55: Why does this logic apply? If there's no name attribute, the resource should be skipped?

bendrucker (commented):
Ok actually I do have a question here. Guessing this is just copy pasted and could be removed.

rules/aws_iam_policy_attachment_exclusive_attachment.go
kayman-mk: I don't see this logic on my machine. We don't need it here.

kayman-mk:
Not sure how to decide about "enabled by default" and "severity".

rules/aws_iam_policy_attachment_exclusive_attachment.go
```diff
@@ -0,0 +1,65 @@
...
+func (r *AwsIAMPolicyAttachmentExclusiveAttachmentRule) Check(runner tflint.Runner) error {
+	resources, err := runner.GetResourceContent(r.resourceType, &hclext.BodySchema{}, nil)
+	if err != nil {
+		return err
+	}
+
+	for _, resource := range resources.Blocks {
+		runner.EmitIssue(
+			r,
+			"Within the entire AWS account, all users, roles, and groups that a single policy is attached to must be specified by a single aws_iam_policy_attachment resource. Consider aws_iam_role_policy_attachment, aws_iam_user_policy_attachment, or aws_iam_group_policy_attachment instead.",
```
wata727, line 55: Personally, I think this explanation is a bit long, but considering that it's disabled by default, it's acceptable.

wata727 (approved):
👍 

> Not sure how to decide about "enabled by default" and "severity".

I believe this rule should remain as is (disabled, warning). Rules should be enabled by default for cases where the issue is worth fixing immediately. The severity should be set to error if the issue will cause `terraform apply` to fail.

PR #851, opened 2025-03-26
provider_missing_default_tags: correctly handle unknown values

### LINKED ISSUE #850: provider_missing_default_tags does not work with dynamic tags
We imports tags from a module. 

And apply them this way : 

```
provider "aws" {
  region = "eu-central-1"
  default_tags {
    tags = module.commons.tags
  }  
}
```

When trying to add the rule : 

```
rule "aws_provider_missing_default_tags" {
    enabled = true
    tags = ["terraform_stack", "terraform_ci_deployed"]
}
```

tflint fails saying that theses tags are not set.

```
❯ tflint --config=./.tflint.hcl --recursive --minimum-failure-severity=notice && echo "succcess"
1 issue(s) found:

Notice: The provider is missing the following tags: "terraform_ci_deployed", "terraform_stack". (aws_provider_missing_default_tags)

  on providers.tf line 27:
  27:     tags = module.commons.tags

Reference: https://github.com/terraform-linters/tflint-ruleset-aws/blob/v0.33.0/docs/rules/aws_provider_missing_default_tags.md
```

If I manually set the tagsm it runs successfuly : 

```
provider "aws" {
  region = "eu-central-1"
  default_tags {
    tags = {
      terraform_ci_deployed = "xoxo"
      terraform_stack       = "fdsfds"
    }
  }
}
```

```
❯ tflint --config=./.tflint.hcl --recursive --minimum-failure-severity=notice && echo "succcess"
succcess
```

I'm not sure if it is expected or not?

bendrucker:
Fixes a bug where `provider_missing_default_tags` does not correctly handle unknown tags. Adds the missing test coverage and then fixes the failing tests by correcting the control flow for unknown values.

### Unknown

The incorrect flow:

```go
if !known {
  return nil
}
```

Doing this _inside_ the `EvaluateExpr` callback effectively ignores the value and leaves `providerTags` empty before checking for unset tags. Instead, unknown values need to short circuit the comparison. So I've returned an error, borrowing `tflint.UnknownError`. 

### Null

The multiple cases where errors are silently discarded was a red flag:

```go
if err != nil {
  return nil
}
```

Now, all errors are returned _except_ evaluation errors of the `tags` attribute. This could be due to an explicit unknown value error or it could be a null value. As in:

```tf
provider "aws" {
  default_tags {
    tags = {
      (var.tag): "bar"
    }
  }
}

variable "tag" {
  type = string
}
```

This is technically not valid configuration, as `var.tag` is `null`, and `null` is not a valid map key. We could just return this error and insist users pass a variable value to TFLint or set a default to make the configuration valid. I initially went that path.

But for the sake of backward compatibility, _any_ error will short circuit checking (as before). But now, the warning log happens _outside_ of the `EvalauteExpr` callback, for _all_ errors, such that they are no longer wholly silent. If the rule is skipping a tag map because of a null key error, it will print that error in a warning.

## Related

Closes #850

### DIFF
--- a/rules/aws_provider_missing_default_tags.go
+++ b/rules/aws_provider_missing_default_tags.go
@@ -1,6 +1,7 @@
 package rules
 
 import (
+	"errors"
 	"fmt"
 	"slices"
 	"sort"
@@ -99,7 +100,7 @@ func (r *AwsProviderMissingDefaultTagsRule) Check(runner tflint.Runner) error {
 				return nil
 			}, nil)
 			if err != nil {
-				return nil
+				return err
 			}
 		}
 		logger.Debug("Walk `%s` provider", providerAlias)
@@ -131,8 +132,7 @@ func (r *AwsProviderMissingDefaultTagsRule) Check(runner tflint.Runner) error {
 				keys, known := getKeysForValue(val)
 
 				if !known {
-					logger.Warn("The missing aws tags rule can only evaluate provided variables, skipping %s.", provider.Labels[0]+"."+providerAlias+"."+providerDefaultTagsBlockName+"."+providerTagsAttributeName)
-					return nil
+					return tflint.ErrUnknownValue
 				}
 
 				logger.Debug("Walk `%s` provider with tags `%v`", providerAlias, keys)
@@ -140,8 +140,13 @@ func (r *AwsProviderMissingDefaultTagsRule) Check(runner tflint.Runner) error {
 				return nil
 			}, nil)
 
+			if errors.Is(err, tflint.ErrUnknownValue) {
+				logger.Warn("The missing aws tags rule can only evaluate provided variables, skipping %s.", provider.Labels[0]+"."+providerAlias+"."+providerDefaultTagsBlockName+"."+providerTagsAttributeName)
+				continue
+			}
+
 			if err != nil {
-				return nil
+				return err
 			}
 
 			// Check tags
--- a/rules/aws_provider_missing_default_tags.go
+++ b/rules/aws_provider_missing_default_tags.go
@@ -140,7 +140,7 @@ func (r *AwsProviderMissingDefaultTagsRule) Check(runner tflint.Runner) error {
 			}, nil)
 
 			if err != nil {
-				logger.Warn("Could not evaluate tags, skipping %s.", provider.Labels[0]+"."+providerAlias+"."+providerDefaultTagsBlockName+"."+providerTagsAttributeName)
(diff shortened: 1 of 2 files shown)


rules/aws_provider_missing_default_tags.go
```diff
@@ -141,7 +140,8 @@ func (r *AwsProviderMissingDefaultTagsRule) Check(runner tflint.Runner) error {
 			}, nil)
 
 			if err != nil {
-				return nil
+				logger.Warn("Could not evaluate tags, skipping %s.%s.%s.%s: %s", provider.Labels[0], providerAlias, providerDefaultTagsBlockName, providerTagsAttributeName, err)
+				continue
```
wata727, line 144: Considering that feedback may be lost in case of simple evaluation errors (e.g. undefined variables), would it be better to only short circuit in the `tflint.ErrUnknownValue` case?

bendrucker: Yes, this is an option, that's what I did originally. Ignoring null variable values would be a separate case. What do you think about erroring there, versus trying to ignore it?

wata727: I think unknown and null values should be ignored, but other errors should not be ignored. How about the following?

```go
if errors.Is(err, tflint.ErrUnknownValue) {
	logger.Warn("Could not evaluate tags, skipping %s.%s.%s.%s: %s", provider.Labels[0], providerAlias, providerDefaultTagsBlockName, providerTagsAttributeName, err)
	continue
}
```

bendrucker: In spirit I agree, but the null case comes in as a raw diagnostic and not a wrapped/simple error:

https://github.com/hashicorp/hcl/blob/b4e1e146a0c3b53f973064ae3fce7b69f5ef13a0/hclsyntax/expression.go#L1219-L1230

wata727: Oops, I missed that. As you say, if the key is null, it doesn't seem to reach `getKeysForValue`.
I misunderstood the following case:
https://github.com/terraform-linters/tflint-ruleset-aws/blob/20dacdce17a538958f8f001ebcac2a2c844733ba/rules/aws_resource_missing_tags.go#L385-L387

wata727 (approved):
👍

rules/aws_provider_missing_default_tags.go
bendrucker: Yeah it's unclear if 100% of that function is actually covered by possible configuration. Particularly this null key case. I'll write some unit tests against it and see if there's dead code.

PR #860, opened 2025-04-09
`aws_write_only_arguments`: recommend write-only arguments where available

### LINKED ISSUE #859: Warn against secrets in state and promote ephemeral resources and values
It would be nice to have linting rules against secrets in state and promote the usage of ephemeral resources and write only values: https://developer.hashicorp.com/terraform/language/resources/ephemeral.

Resources that have ephemeral equivalents:
- [`ssm_parameter`](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/ephemeral-resources/ssm_parameter)
- [`secretsmanager_secret_version `](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/ephemeral-resources/secretsmanager_secret_version)
- [`secretsmanager_random_password `](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/ephemeral-resources/secretsmanager_random_password)
- [`eks_cluster_auth`](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/ephemeral-resources/eks_cluster_auth)
- [`kms_secrets`](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/ephemeral-resources/kms_secrets)

Resources with write-only attributes:
- [`rds_cluster`](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/rds_cluster)
- [`redshift_cluster`](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/redshift_cluster)
- [`docdb_cluster`](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/docdb_cluster)
- [`redshiftserverless_namespace`](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/redshiftserverless_namespace)
- [`ssm_parameter`](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/ssm_parameter)

aristosvo:
Fixes #859 partially

A review on the direction of these kind of checks would be really appreciated, I would happily implement all the write-only arguments once we have a direction.


bendrucker (commented):
I would suggest tying this up into one meta-rule for recommending ephemeral resources and write only. Or two. Versus doing this on a per-resource basis.

Then you can just drive it off configuration/data and quickly handle all of them without all boilerplate. E.g. for write only attributes, a map of resource to a list of attributes with write only alternatives: 

```go
type writeOnlyAttribute struct {
    original    string
    alternative string
}

writeOnlyAttributes := map[string][]writeOnlyAttribute{
    "aws_secretsmanager_secret_version": {
        {
            original:    "secret_string",
            alternative: "secret_string_wo",
        },
    },
}
```

Additionally, you should use `EmitIssueWithFix` here so that users can automatically convert their code with `--fix`. There are no existing autofixes in this repository but you can see an example here:

https://github.com/terraform-linters/tflint-ruleset-terraform/blob/c2e1c2e3337034dc475f77d574de64213384c2b4/rules/terraform_deprecated_index.go#L80

Here's the methods you can call to make replacements:

https://pkg.go.dev/github.com/terraform-linters/tflint-plugin-sdk/tflint#Fixer

aristosvo:
Thanks @bendrucker ! Based on your suggestions it is implemented for the write-only attributes.

I’ll have to puzzle a bit how to recommend using an ephemeral resource instead of a data source, but I think that would be a different MR.

I didn’t include the automatic version checking (yet), as that raised a lot of extra questions like:
- if they are on a version lower that 1.11, should I raise this issue at all? 
- if raised, should I recommend them to update, maybe even auto fix that?
- can I make check opt-in for versions lower than 1.11, and opt-out for versions 1.11 and later?

Sorry for all the questions, but it seems this rule has some unique patterns to it 😅

docs/rules/aws_write_only_attributes.md
```diff
@@ -0,0 +1,57 @@
...
+Warning: [Fixable] "secret_string" is a non-ephemeral attribute, which means this secret is stored in state. Please use "secret_string_wo". (aws_write_only_attributes)
+
+  on test.tf line 3:
+   3:   secret_string = var.secret
+
+```
+
+## Why
+
+Saving secrets to state or plan files is a bad practice. It can cause serious security issues. Keeping secrets from these files is possible in most of the cases by using write-only attributes.
```
bendrucker, line 28:
```suggestion
By default, sensitive attributes are still stored in state, just hidden from view in plan output. Other resources are able to refer to these attributes. Current versions of Terraform also include support for write-only arguments, which are not persisted to state. Other resources cannot refer to their values.

Using write-only arguments mitigates the risk of a malicious actor obtaining privileged credentials by accessing Terraform state files directly. Prefer using them over the original sensitive attribute unless you need to refer to it in other blocks, such as a [root `output`](https://developer.hashicorp.com/terraform/language/values/outputs#ephemeral-avoid-storing-values-in-state-or-plan-files), that cannot be ephemeral.
```

Security recommendations should be precise, on:

* The risk/threat
* The mitigation
* Any tradeoffs/limitations

Let the reader interpret how it applies to their code/system. Avoid subjective characterization, e.g.,:

* Bad practice
* "Serious" issues
* "possible in most of the cases"

docs/rules/aws_write_only_attributes.md
```diff
@@ -0,0 +1,57 @@
+# aws_write_only_attributes
+
+Disallows using the normal attribute containing sensitive information, in favor of an available write-only attribute. This is only valid for Terraform version 1.11.0 and later. If there alternative options available to prevent this information ending up in state, these are suggested as well.
+
+## Example
+
+In this example `aws_secretsmanager_secret_version` is used, but you can replace that with any resource with write-only attributes:
```
bendrucker, line 7:
```suggestion
This example uses `aws_secretsmanager_secret_version`, but the rule applies to all resources with write-only attributes:
```

docs/rules/aws_write_only_attributes.md
```diff
@@ -0,0 +1,57 @@
+# aws_write_only_attributes
+
+Disallows using the normal attribute containing sensitive information, in favor of an available write-only attribute. This is only valid for Terraform version 1.11.0 and later. If there alternative options available to prevent this information ending up in state, these are suggested as well.
```
bendrucker, line 3:
```suggestion
Recommends using available [write-only arguments](https://developer.hashicorp.com/terraform/language/resources/ephemeral/write-only) instead of the original sensitive attribute. This is only valid for Terraform v1.11+.
```

docs/rules/aws_write_only_attributes.md
```diff
@@ -0,0 +1,57 @@
...
+
+```
+
+## Why
+
+Saving secrets to state or plan files is a bad practice. It can cause serious security issues. Keeping secrets from these files is possible in most of the cases by using write-only attributes.
+
+## How To Fix
+
+Replace the `secret_string` with `secret_string_wo`, preferable via using an ephemeral resource as source or using an ephemeral variable:
```
bendrucker, line 32:
```suggestion
Replace the attribute with its write-only argument equivalent. Reference an ephemeral resource or ephemeral variable to ensure that the sensitive value is not persisted to state.
```

rules/aws_write_only_attributes.go
```diff
@@ -0,0 +1,127 @@
...
+type AwsWriteOnlyAttributesRule struct {
+	tflint.DefaultRule
+
+	writeOnlyAttributes map[string]writeOnlyAttribute
+}
+
+type writeOnlyAttribute struct {
+	original             string
+	writeOnlyAlternative string
+	otherAlternative     string
```
bendrucker, line 23: Leave this out, it is too big a change recommendation to offer without context. 

Adopting the write-only alternative to a sensitive attribute has clear effects that are documented in the rule:

1. Avoids inclusion in state
2. Prevents reference from other blocks

The `manage_*` attributes are a substantially different feature, opting the user into a new secret in Secrets Manager and notably _automatic rotation_. 

This deserves its own rule, even if both changes share the effect 1.

rules/aws_write_only_attributes.go
```diff
@@ -0,0 +1,127 @@
...
+}
+
+// NewAwsWriteOnlyAttributesRule returns new rule with default attributes
+func NewAwsWriteOnlyAttributesRule() *AwsWriteOnlyAttributesRule {
+	writeOnlyAttributes := map[string]writeOnlyAttribute{
+		"aws_secretsmanager_secret_version": {
```
bendrucker, line 29: On first glance I notice `aws_db_instance` missing and that probably means plenty of other less common services are missing too.

You could generate these:

https://developer.hashicorp.com/terraform/cli/commands/providers/schema

Go through all resources looking for `*_wo`. Even if you don't actually add a generator here just doing it once at least guarantees your initial data is complete.

bendrucker (commented):
Nice, few more edits, but getting close. 

Also, it'll be annoying, but in reviewing the docs:

https://developer.hashicorp.com/terraform/language/resources/ephemeral/write-only

They decided to call these write-only _arguments_ and not attributes, presumably to emphasize that they cannot be referenced like traditional attributes. I know I called them by the wrong name initially/in my example, but we do need to match Terraform's nomenclature.

rules/aws_write_only_attributes.go
aristosvo: I've generated both rule and test based on the schema 💪

Thanks for showing the way 🚀

rules/aws_write_only_attributes.go
```diff
@@ -0,0 +1,127 @@
...
+}
+
+// Name returns the rule name
+func (r *AwsWriteOnlyAttributesRule) Name() string {
+	return "aws_write_only_attributes"
+}
+
+// Enabled returns whether the rule is enabled by default
+func (r *AwsWriteOnlyAttributesRule) Enabled() bool {
+	return true
```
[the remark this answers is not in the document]
aristosvo: I made the rule optional for now, seems like the best way to go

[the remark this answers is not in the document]
wata727: Opened a related feature request to the SDK https://github.com/terraform-linters/tflint-plugin-sdk/issues/391

rules/ephemeral/aws_write_only_arguments.go
```diff
@@ -0,0 +1,139 @@
...
+
+			for _, resource := range resources.Blocks {
+				attribute, exists := resource.Body.Attributes[resourceAttribute.originalAttribute]
+				if !exists {
+					continue
+				}
+
+				err := runner.EvaluateExpr(attribute.Expr, func(val cty.Value) error {
+					if !val.IsNull() {
+						if err := runner.EmitIssueWithFix(
```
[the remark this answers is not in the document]
aristosvo: I agree. Would you recommend auto fixing these version attributes as well or remove the autofix completely?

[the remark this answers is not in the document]
wata727: Personally, I would recommend removing the autofix, to suggest that users should manually fix in context.
@bendrucker What do you think about this?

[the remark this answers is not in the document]
bendrucker: Why does `secret_string_wo_version` need to be set? I think auto-fixing makes sense even if there's upstream work to make references ephemeral. Another rule can continue to warn about those.

But if there's a need to make other changes as part of write-only adoption, then auto fixing might be confusing.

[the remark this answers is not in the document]
wata727: When setting `secret_string_wo`, you are forced to set `secret_string_wo_version` by `RequiredWith`.
https://github.com/hashicorp/terraform-provider-aws/blob/v5.94.1/internal/service/secretsmanager/secret_version.go#L85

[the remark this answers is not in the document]
bendrucker: Ok in that case the rule should identify that other attribute. Autofixing it in would be nice but if it's too hard to get a correct fix we can start without.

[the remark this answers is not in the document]
aristosvo: I will give it a short try, and otherwise remove the auto fix

bendrucker (commented):
Nice work on the auto-fix!

Last thing: this rule isn't an ideal candidate for a generator, at least for the rule code itself. You need a generator when you want to build multiple rules from data. You can't create types in a loop. 

But in this case you just have one rule, and could have your generator just write the relevant data for the rule to use (e.g., a JSON file) in its `Check` method. You'd [embed](https://pkg.go.dev/embed) the file. Or, you could generate Go code with a global variable. 

Either way, this means you can test your rule for a handful of representative cases rather than having generated tests for every resource.

aristosvo:
> Nice work on the auto-fix!
> 
> Last thing: this rule isn't an ideal candidate for a generator, at least for the rule code itself. You need a generator when you want to build multiple rules from data. You can't create types in a loop.
> 
> But in this case you just have one rule, and could have your generator just write the relevant data for the rule to use (e.g., a JSON file) in its `Check` method. You'd [embed](https://pkg.go.dev/embed) the file. Or, you could generate Go code with a global variable.
> 
> Either way, this means you can test your rule for a handful of representative cases rather than having generated tests for every resource.

Thanks @bendrucker!

- [x] I generated a global variable instead of the complete rule, reduced tests to on type of resource instead of all of them. Embedding a JSON file felt a bit off compared to the global variable, although that might be a taste issue 🙃

wata727 (approved):
Looks good to me. Will merge after @bendrucker's approval.

rules/ephemeral/aws_write_only_arguments_test.go
```diff
@@ -0,0 +1,65 @@
...
+	rule := NewAwsWriteOnlyArgumentsRule()
+
+	for _, tc := range cases {
+		filename := "resource.tf"
+		runner := helper.TestRunner(t, map[string]string{filename: tc.Content})
+
+		if err := rule.Check(runner); err != nil {
+			t.Fatalf("Unexpected error occurred: %s", err)
+		}
+		helper.AssertIssuesWithoutRange(t, tc.Expected, runner.Issues)
```
bendrucker, line 57: It would be nice to use plain `AssertIssues` to ensure that the range is properly set to the `secret_string` attribute. Perhaps this is a leftover from using generation where that was changing depending on the resource, as the key length would differ.

rules/ephemeral/generator/main.go
```diff
@@ -0,0 +1,68 @@
+//go:build generators
```
bendrucker, line 1: This is not necessary since the generator is a different package. When `go build` runs, it walks the graph of all `import` statements found in the package for which it was invoked, plus transitive dependencies. This generator shouldn't be anywhere in that graph, since it never actually gets imported, only invoked via `go run`. You'd only need a build tag on this file if the package were imported in order to use things exported from other files.

bendrucker (approved):
Nice! Two small comments, neither essential.

rules/ephemeral/aws_write_only_arguments_test.go
aristosvo: Thanks. 

✅ Resolved!

bendrucker:
Thanks again!

PR #861, opened 2025-04-12
feat: warn against data sources with ephemeral alternatives

aristosvo:
Warns against data sources with ephemeral alternatives to prevent storage of sensitive data in state as much as possible.

Follow up of #860


rules/ephemeral/aws_ephemeral_resources.go
```diff
@@ -0,0 +1,90 @@
...
+			return err
+		}
+
+		for _, resource := range resources.Blocks {
+			if err := runner.EmitIssueWithFix(
+				r,
+				fmt.Sprintf("\"%s\" is a non-ephemeral data source, which means that all (sensitive) attributes are stored in state. Please use ephemeral resource \"%s\" instead.", resourceType, resourceType),
+				resource.TypeRange,
+				func(f tflint.Fixer) error {
+					return f.ReplaceText(resource.TypeRange, "ephemeral")
```
wata727, line 59: I'm wondering if this autofix is ​​safe. Expressions that reference the data attribute must also be rewritten to `ephemeral.*`, otherwise an error will occur. Furthermore, ephemeral attributes can only be referenced in certain places, in which case only data resources are valid.

Personally, I find this autocorrect too complicated and prefer to start without it.

docs/rules/aws_ephemeral_resources.md
wata727, line 1: Can you add a link to this documentation in `docs/rules/README.md`?
Please also add the link to `aws_write_only_attributes` rule.

rules/ephemeral/aws_ephemeral_resources.go
aristosvo: Thanks, I'll remove it.

I didn't know exactly what the qualifications are for auto-fix, but based on this and previous contribution it seems that auto-fix can only be applied if it really fixes all loose ends. That is indeed not the case here, it is only a starting point.

docs/rules/README.md
```diff
@@ -80,6 +80,15 @@ These rules enforce best practices and naming conventions:
...
 |[aws_provider_missing_default_tags](aws_provider_missing_default_tags.md)|Require specific tags for all AWS providers default tags||
 
+### Removing secrets from state
+
+These rules recommend best practices to keep sensitive information from state:
+
+|Rule|Description|Enabled by default|
+| --- | --- | --- |
+|[aws_ephemeral_resources](aws_ephemeral_resources.md)|Recommends using available ephemeral resources instead of the original data source. This is only valid for Terraform v1.10+.||
+|[aws_write_only_arguments](aws_write_only_arguments.md)|Recommends using available write-only arguments instead of the original sensitive attribute. This is only valid for Terraform v1.11+.||
```
wata727, line 90: I believe these should be added to the "Best Practices" section.

docs/rules/aws_ephemeral_resources.md
```diff
@@ -0,0 +1,44 @@
...
+  password_length = 50
+  exclude_numbers = true
+}
+```
+
+```
+$ tflint
+1 issue(s) found:
+
+Warning: [Fixable] "aws_secretsmanager_random_password" is a non-ephemeral data source, which means that all (sensitive) attributes are stored in state. Please use ephemeral resource "aws_secretsmanager_random_password" instead. (aws_ephemeral_resources)
```
wata727, line 20: This rule is currently not fixable.

docs/rules/README.md
aristosvo: Thanks, fixed ✅

docs/rules/aws_ephemeral_resources.md
aristosvo: Done⚡

wata727 (approved):
Thanks!

PR #1020, opened 2025-12-20
Generate validation rules for map attributes

bendrucker:
Generates validation rules for `tags` and other `map(string)` attributes by extracting key/value constraints from AWS Smithy models. This catches invalid tags at plan time rather than waiting for API errors at apply.

For example, given:

```hcl
resource "aws_ecs_service" "example" {
  tags = {
    ""             = "value"   # key must be at least 1 character
    "invalid!key"  = "value"   # key contains invalid character "!"
    "aws:reserved" = "value"   # "aws:" prefix is reserved (select services)
    "Name"         = "x<y>"    # value contains invalid characters
  }
}
```

The rule reports each violation with the specific constraint (pattern, length, or prefix) that failed. The same approach applies to other map attributes like `labels`, `parameters`, `attributes`, and `default_run_properties`.

In total, this adds 158 rules: 146 tag rules across 48 services, plus 12 rules for other map attributes (EKS labels, Glue parameters, X-Ray attributes, etc.).

## Changes

- Adds `map(ItemsShape, KeyShape, ValueShape)` function to mapping file syntax for services that define tags as list-of-structure types
- Shapes in eval context are rich objects carrying all constraint info (pattern, length, enum)
- Transform functions (`uppercase`, `replace`) operate on shape enum values
- Adds a handwritten `mapRule` type holding the validation logic, and generates every map rule as a constraint literal in one `map_rules.go` table
- Validates max tag count per resource using Smithy length traits (e.g., 50 for ECS, 200 for S3)
- Extracts prefix deny patterns from negative lookaheads (e.g., `(?!aws:)`) and generates `strings.HasPrefix` checks, since Go's regexp doesn't support lookahead
- Skips match-all patterns and Smithy sentinel max values (`2147483647`) to avoid generating no-op checks
- Uses attribute name in error messages (e.g., "definition key" for non-tag map rules)
- Counts key and value length in characters, matching how AWS states tag limits
- Formats generated Go files with `go/format` inside the generator, so templates no longer need to emit gofmt-clean text

## Implementation

AWS services define tags using two different patterns in their Smithy models.

#### Smithy `map` Type

Some services use a `map` type with explicit `key` and `value` targets:

```
TagMap (map)
  ├── key → TagKey (string with constraints)
  └── value → TagValue (string with constraints)
```

The generator traverses this automatically via `traverseToMapConstraints()`, so these services keep their existing mapping entries.

#### List of Structures

Most services use a list containing a structure with named members:

```
TagList (list)
  └── member → Tag (structure)
                 ├── Key → TagKey (string with constraints)
                 └── Value → TagValue (string with constraints)
```

This pattern cannot be inferred automatically because member names vary across services. The `map()` function explicitly specifies which shapes provide items, key, and value constraints:

```hcl
tags = map(Tags, TagKey, TagValue)
```

### Prefix Deny Patterns

Tag key patterns often include negative lookahead like `(?!aws:)` to prevent reserved prefixes. Since Go's regexp doesn't support lookahead, the generator extracts literal prefix strings and emits `strings.HasPrefix` checks in the template.

- A single literal prefix becomes one check.
- An alternation of literal prefixes such as `(?!aws:|connect:)` becomes one check per prefix.
- A lookahead the generator cannot convert is reported as a warning during generation.

## Testing

Table tests exercise the shared `mapRule` type directly: check ordering, character counting, sensitive messages, and singular/plural wording. Three tests run representative generated rules end to end through the runner, each exercising a distinct constraint shape:

- **ECS**: full constraints, covering items max, key/value pattern and length, and invalid characters
- **Signer**: prefix deny (`aws:`) combined with key pattern validation
- **S3**: minimal constraints, with key min length only and no pattern or value checks

Generator-level tests cover shape traversal (map, list, structure types), `replacePattern` transforms (unicode escapes, lookahead extraction, anchors), and shape object composition.

## References

- Closes #1019


wata727 (approved):
Nice!

PR #1084, opened 2026-04-19
Add FreeBSD build target

adyxax:
Thank you for making tflint and its very useful rulesets!

FreeBSD packages `tflint`, but none of the useful plugins seem to be released for this operating system yet. This very small patch adds `freebsd_amd64` as a build target.

If testing before new releases is a concern, I volunteer to do it. I am using `tflint` in all my tofu projects and using FreeBSD a lot. If this contribution is well received, I will happily submit the same change to the azurerm ruleset too.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -7,6 +7,7 @@ builds:
   - targets:
       - darwin_amd64
       - darwin_arm64
+      - freebsd_amd64
       - linux_386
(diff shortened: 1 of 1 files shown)


wata727:
First of all, TFLint itself doesn't support FreeBSD build target, so is there any reason why it would be good to have FreeBSD support only in this ruleset plugin?

PR #1091, opened 2026-04-30
fix: avoid expanding unrelated resources in IAM policy SID rule

### LINKED ISSUE #1085: Flaky error - failed to check "aws_iam_policy_sid_invalid_characters"
### Issue

This issue seems related to https://github.com/terraform-linters/tflint-ruleset-aws/issues/749.

Error:
```
16:39:21 [TRACE] interceptor/logging.go:15: gRPC request: direction=plugin2host method=/proto.Runner/GetModuleContent req="schema:{blocks:{type:\"resource\"  label_names:\"type\"  label_names:\"name\"  body:{attributes:{name:\"policy\"}  Mode:SCHEMA_MODE_DEFAULT}}  Mode:SCHEMA_MODE_DEFAULT}  option:{module_ctx:MODULE_CTX_TYPE_SELF  hint:{resource_type:\"aws_iam_policy\"}  expand_mode:EXPAND_MODE_EXPAND}"
16:39:21 [ERROR] interceptor/logging.go:18: failed to gRPC request: direction=plugin2host method=/proto.Runner/GetModuleContent err="rpc error: code = FailedPrecondition desc = local.tf:97,28-42: Unsupported attribute; This object does not have an attribute named \"lb_controller\"."
16:39:21 [DEBUG] go-plugin@v1.7.0/client.go:1246: tflint-ruleset-aws: 16:39:21 [ERROR] interceptor/logging.go:18: failed to gRPC request: direction=host2plugin method=/proto.RuleSet/Check err="rpc error: code = Aborted desc = failed to check \"aws_iam_policy_sid_invalid_characters\" rule: local.tf:97,28-42: Unsupported attribute; This object does not have an attribute named \"lb_controller\"."
```

`local.tf` file:
```
locals {
...
  ingresses = [
    ..
    local.namespaces.system.ingress_internal,
    local.namespaces.system.lb_controller     <-- line 97
  ]
```

It' just a local declaration, not even an `aws_iam_policy` resource. It's really weird because if I remove non-related content from other files or other sections it works just fine. I'll try to get a mininum setup to reproduce it.

### Versions

```
TFLint version 0.62.0
+ ruleset.terraform (0.14.1-bundled)
```

Config:
```
plugin "aws" {
  enabled    = true
  deep_check = false
  source     = "github.com/terraform-linters/tflint-ruleset-aws"
  version    = "0.47.0"
}
```

dbuaon:
## Problem

`aws_iam_policy_sid_invalid_characters` could fail before evaluating an `aws_iam_policy` resource when unrelated resources in the same module contained expressions that could not be expanded.

In #1085, an unrelated `kubernetes_network_policy` used `for_each = toset(local.ingresses)`, where `local.ingresses` referenced an attribute that does not exist with the default variable values.

## Root cause

The rule retrieved `aws_iam_policy` resources using the default expansion behavior. This caused TFLint to evaluate unrelated resource expressions during `GetModuleContent`.

## Fix

Use `ExpandModeNone` when retrieving `aws_iam_policy` resources, because this rule only needs to inspect the `policy` attribute and does not require resource expansion.

## Validation

- Reproduced the issue locally with the published AWS ruleset.
- Verified the fix locally using `plugin_dir` and a locally built plugin.
- Added an integration regression fixture for #1085.
- Ran:

```bash
go test ./rules -run Test_AwsIAMPolicySidInvalidCharacters -count=1
go test ./integration -run 'TestIntegration/unrelated_expansion_failure_with_--only_aws_iam_policy_sid_invalid_characters' -count=1
go test ./... -count=1

Fixes #1085


wata727:
This change is not appropriate because calling `GetModuleContent` with `ExpandModeNone` will also prevent `aws_iam_policy` from being expanded. For example, if the `policy` attribute has `count.index` or `each.key`, it will be ignored.

dbuaon:
Thanks, that makes sense.

I understand why `ExpandModeNone` is not appropriate here: it avoids expansion of unrelated resources, but it also prevents expansion of the target `aws_iam_policy`, which can break valid cases involving `count.index`, `each.key`, or other expressions in the `policy` attribute.

Before closing this PR, I want to confirm whether there is another preferred direction.

Would it make sense to move the investigation to TFLint core, where `GetModuleContent` could potentially avoid expanding unrelated resources, or is this considered a known limitation of the current HCL/TFLint expansion model?

If there is no safe way to preserve target resource expansion while avoiding unrelated resource expansion, I can close this PR.

wata727:
I'm closing this PR because I don't think it should be fixed in the ruleset.
While it's possible that this could be fixed in the TFLint core, as explained at https://github.com/terraform-linters/tflint-ruleset-aws/issues/1085#issuecomment-4350501639, it's a design in the HCL API, so I'm not sure if a reasonable workaround can be found.

PR #1108, opened 2026-06-08
Check cloudwatch retention best practice

eldondevat:


bendrucker:
> Many compliance frameworks

Compliance guardrails are not a good fit for this repo. This isn't really a best practice for using an AWS service it's just a common requirement that many organizations might have.

bendrucker:
And by common I mean that it's not unusual but it's not straightforward either. SOC2 certainly doesn't specify 365, even if that's "common" too as a default among auditors. Style and compliance preferences just turn into sources of churn, best to keep these types of rules within your own org.

eldondevat:
Thanks for the prompt response!

PR #1160, opened 2026-08-28
Add r8g, m8g and c8gn ElastiCache node types

StevenKearns:
Both `invalid_type` rules currently reject 21 node types AWS sells. 

I added the `cache.r8g`, `cache.m8g` and `cache.c8gn` families. 

Verified against the `AmazonElastiCache` offer file in `us-east-1`, `us-west-2`, `eu-west-1` and `ap-northeast-1` - the size sets are identical and all 21 are current generation.


bendrucker:
Thanks, superseded by #1163, which generates the node types from the AWS Price List and the ElastiCache API model rather than continuing maintain the list by hand. It includes the types added here.

